PR #201, opened 2026-06-25
Preserve diff line locations in SARIF output

bomly-guy:
## Summary
- Pass changed git line ranges through diff resolution into SARIF output
- Attach multiple declaration-site positions for detectors so package locations stay accurate across manifests and versions
- Expand detector tests to cover version-aware and duplicate-location cases

## Testing
- Not run (not requested)

## Summary by CodeRabbit

* **New Features**
  * SARIF results now prioritize findings that overlap changed lines, making diffs more relevant.
  * Dependency locations are now captured more precisely across several ecosystems, including multiple matching source locations when needed.

* **Bug Fixes**
  * Improved line-number accuracy for Python, Maven, npm, pnpm, NuGet, Conan, SwiftPM, Pub, GitHub Actions, and Git diffs.
  * Lockfile-based scans now include the same location details as other resolution paths.

* **Tests**
  * Added coverage for changed-line SARIF behavior and updated detector position expectations.

### DIFF
--- a/internal/detectors/conan/positions.go
+++ b/internal/detectors/conan/positions.go
@@ -10,6 +10,7 @@ import (
 )
 
 var conanRefLine = regexp.MustCompile(`([a-zA-Z0-9_][a-zA-Z0-9._+-]*)\s*/\s*([^@'"\s,)]+)`)
+var conanPythonRequireLine = regexp.MustCompile(`(?:self\.)?(?:requires|build_requires|tool_requires|test_requires)\s*\(\s*["']([a-zA-Z0-9_][a-zA-Z0-9._+-]*)\s*/\s*([^@'"\s,)]+)`)
 
 var conanSectionHeader = regexp.MustCompile(`^\s*\[\s*(requires|build_requires|tool_requires|test_requires)\s*\]\s*$`)
 
@@ -33,6 +34,9 @@ func conanPositions(projectDir string) map[string][]*sdk.SourcePosition {
 				return
 			}
 			matches := conanRefLine.FindStringSubmatch(text)
+			if name == "conanfile.py" {
+				matches = conanPythonRequireLine.FindStringSubmatch(text)
+			}
 			if matches == nil {
 				return
 			}
@@ -42,9 +46,9 @@ func conanPositions(projectDir string) map[string][]*sdk.SourcePosition {
 			}
 			version := strings.TrimSpace(matches[2])
 			pos := &sdk.SourcePosition{File: name, Line: line}
-			appendPosition(out, pkgName, pos)
+			detectors.AppendPosition(out, pkgName, pos)
 			if version != "" {
-				appendPosition(out, pkgName+"@"+version, pos)
+				detectors.AppendPosition(out, pkgName+"@"+version, pos)
 			}
 		})
 	}
@@ -70,17 +74,4 @@ func AttachConanPositions(g *sdk.Graph, projectDir string) {
 		}
 		return []string{name + "@" + strings.TrimSpace(pkg.Version), name}
 	})
-}
-
-func appendPosition(out map[string][]*sdk.SourcePosition, key string, pos *sdk.SourcePosition) {
-	key = strings.TrimSpace(key)
-	if key == "" || pos == nil {
-		return
-	}
-	for _, existing := range out[key] {
-		if existing.File == pos.File && existing.Line == pos.Line && existing.Column == pos.Column {
-			return
-		}
-	}
-	out[key] = append(out[key], pos)
 }
--- a/internal/detectors/conan/positions.go
+++ b/internal/detectors/conan/positions.go
@@ -9,20 +9,16 @@ import (
 	"github.com/bomly-dev/bomly-cli/sdk"
 )
 
-// conanRequireLine matches a Conan require line of the form
-// `name/version[@user/channel]` typically found in conanfile.txt's
-// [requires] / [build_requires] / [tool_requires] sections. The
-// captured value is the package name.
-var conanRequireLine = regexp.MustCompile(`^\s*([a-zA-Z0-9_][a-zA-Z0-9._+-]*)\s*/\s*[^@\s]+`)
+var conanRefLine = regexp.MustCompile(`([a-zA-Z0-9_][a-zA-Z0-9._+-]*)\s*/\s*([^@'"\s,)]+)`)
 
 var conanSectionHeader = regexp.MustCompile(`^\s*\[\s*(requires|build_requires|tool_requires|test_requires)\s*\]\s*$`)
 
-func conanPositions(projectDir string) map[string]*sdk.SourcePosition {
-	out := make(map[string]*sdk.SourcePosition)
-	candidates := []string{"conanfile.txt", "conan.lock", "conaninfo.txt"}
+func conanPositions(projectDir string) map[string][]*sdk.SourcePosition {
+	out := make(map[string][]*sdk.SourcePosition)
+	candidates := []string{"conanfile.txt", "conanfile.py", "conan.lock", "conaninfo.txt"}
 	for _, name := range candidates {
 		full := filepath.Join(projectDir, name)
-		insideRequires := name == "conan.lock" || name == "conaninfo.txt" // always scan lock/info bodies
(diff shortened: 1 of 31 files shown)


internal/output/sarif.go
```diff
@@ -426,6 +460,40 @@ func sarifLocationsForFinding(f sdk.Finding, includeReachability bool, options [
...
 }
 
+func sarifChangedLines(options []SARIFOptions) map[string][]SARIFLineRange {
+	if len(options) == 0 || len(options[0].ChangedLines) == 0 {
+		return nil
+	}
+	return options[0].ChangedLines
+}
+
+func sarifLocationDiffScore(uri string, region *sarifRegion, changedLines map[string][]SARIFLineRange) int {
```
bomly-guy, line 478: **Limitation worth documenting: this only works for repos whose project is at the git root.** `Position.File` is set relative to each detector's `workingDir`/`ProjectPath` (e.g. `package-lock.json`, `pom.xml`), and nothing rebases it to a repo-root-relative path after detection. The `changedLines` keys here come from `git diff` and are repo-root-relative (e.g. `apps/web/package-lock.json`).

So for any subproject **not at the repo root**: (a) `changedLines[uri]` never matches → every candidate scores 2 → no prioritization, and (b) the emitted SARIF URI itself is wrong (points at a root-level file that may not exist). This is pre-existing (the PR doesn't regress it) and your single-root dogfood repos are fine, but multi-module Maven / monorepo layouts won't benefit and may mis-attribute. Worth a `dev-docs/ARCHITECTURE.md` note and a follow-up to prefix `Position.File` with the subproject `RelativePath` during consolidation.

internal/output/sarif.go
```diff
@@ -397,16 +411,36 @@ func sarifLocationsForFinding(f sdk.Finding, includeReachability bool, options [
...
+			}
+		}
+		locations := make([]sarifLocation, 0, len(candidates))
+		for _, candidate := range candidates {
+			if len(changedLines) > 0 && candidate.score != bestScore {
+				continue
+			}
+			locations = append(locations, candidate.location)
+		}
 		return locations, originalURIs
```
bomly-guy: **Scoring is global per-finding, not per-dependency.** All candidate locations across *every* dependency in `f.DependencyRefs` are pooled, then everything not equal to the single `bestScore` is dropped. For a finding that references multiple distinct packages (e.g. a policy finding) where only one was touched by the PR, the locations for the *untouched* packages get pruned entirely.

For per-package vuln findings (the common case) this is harmless, but it's a latent over-pruning bug. Consider computing `bestScore` and filtering **per dependency** rather than across the whole finding.

internal/output/sarif.go
```diff
@@ -159,6 +159,14 @@ type SARIFOptions struct {
 	IncludeReachability bool
 	LocationGraphs      []*sdk.Graph
 	BaselineState       string
+	ChangedLines        map[string][]SARIFLineRange
```
bomly-guy, line 162: Confirming for the checklist: `ChangedLines`/`SARIFLineRange` are function-option fields, not part of the emitted SARIF JSON or config schema, so `make generate` is **not** required here despite the CLAUDE.md "if `internal/output/*` changes" rule. No action needed — just flagging so a reviewer doesn't ask for regenerated docs.

internal/detectors/conan/positions.go
```diff
@@ -62,10 +60,27 @@ func AttachConanPositions(g *sdk.Graph, projectDir string) {
...
+		name := strings.TrimSpace(pkg.Name)
+		if name == "" {
+			return nil
 		}
-		return strings.TrimSpace(pkg.Name)
+		return []string{name + "@" + strings.TrimSpace(pkg.Version), name}
 	})
 }
+
+func appendPosition(out map[string][]*sdk.SourcePosition, key string, pos *sdk.SourcePosition) {
```
bomly-guy: This exact `appendPosition` helper is now copy-pasted verbatim into 6 packages (conan, githubactions, maven, nuget, node/npm, node/pnpm). CLAUDE.md prefers shared helpers in `internal/detectors`. Suggest hoisting it next to `AttachPositionCandidates` as an exported `detectors.AppendPosition(out, key, pos)` and deleting the six copies.

internal/detectors/conan/positions.go
```diff
@@ -9,20 +9,16 @@ import (
 	"github.com/bomly-dev/bomly-cli/sdk"
 )
 
-// conanRequireLine matches a Conan require line of the form
-// `name/version[@user/channel]` typically found in conanfile.txt's
-// [requires] / [build_requires] / [tool_requires] sections. The
-// captured value is the package name.
-var conanRequireLine = regexp.MustCompile(`^\s*([a-zA-Z0-9_][a-zA-Z0-9._+-]*)\s*/\s*[^@\s]+`)
+var conanRefLine = regexp.MustCompile(`([a-zA-Z0-9_][a-zA-Z0-9._+-]*)\s*/\s*([^@'"\s,)]+)`)
```
bomly-guy, line 12: `conanRefLine` is unanchored (no `^`) and is now run over the **entire** `conanfile.py` body (`insideRequires` is forced true for `.py` on line 21). It will extract spurious `name/version` pairs from URLs/paths/comments — e.g. a line containing `github.com/org/repo` yields a bogus `github.com` → `org/repo` match. It's harmless to output today (no graph node matches the noise), but a comment like `# bump foo/1.2.3` where `foo` is a real dep could attach a wrong line. Consider anchoring to `requires`/`self.requires(...)`-style call sites for `.py`, or at least add a comment noting the best-effort, noisy nature of the `.py` scan.

internal/detectors/swiftpm/positions.go
```diff
@@ -12,23 +12,35 @@ import (
...
+		if _, exists := out[pendingName]; !exists {
+			out[pendingName] = &sdk.SourcePosition{File: relPath, Line: line}
+		}
+		pendingName = ""
+		pendingLine = 0
 	})
+	if pendingName != "" {
+		if _, exists := out[pendingName]; !exists {
+			out[pendingName] = &sdk.SourcePosition{File: relPath, Line: pendingLine}
+		}
```
bomly-guy: Two asymmetries vs. the in-loop path here:

1. In-loop records the **version/revision** line (line 34); this end-of-file flush records `pendingLine`, the **identity/name** line. The fallback should also point at the version line for consistency (or carry a `pendingVersionLine`).
2. More importantly, an `identity` block with no `version`/`revision` before the next `identity` is **silently dropped** mid-stream (the next identity overwrites `pendingName` without recording), so only the very last such block survives via this flush. Previously every identity was recorded. Every resolved entry normally has a revision so this is an edge case, but worth a guard/comment.

bomly-guy (commented):
Reviewed end-to-end (full diff, root causes in base, built + ran `detectors/...`, `output`, `git`, `cli` on this head — all green; the PR description said testing wasn't run). The two-pronged design is right: capture multiple declaration sites per package keyed by `name@version`/`name`, then let the SARIF diff-scorer keep only the diff-relevant one. This correctly addresses all the dogfooding cases — Python lock fast-paths now attach positions (the pipenv path already did, which is why pipenv worked), npm/pnpm disambiguate by version, and GitHub Actions no longer pins an out-of-diff workflow.

Leaving inline notes: two substantive (subproject paths, per-finding scoping) and several minor cleanups. None are blockers.

internal/detectors/pub/positions.go
```diff
@@ -29,19 +32,28 @@ func pubspecLockPositions(path, relPath string) map[string]*sdk.SourcePosition {
...
-		out[name] = &sdk.SourcePosition{File: relPath, Line: line}
+		if _, exists := out[pendingName]; !exists {
+			out[pendingName] = &sdk.SourcePosition{File: relPath, Line: line}
+		}
+		pendingName = ""
+		pendingLine = 0
 	})
+	if pendingName != "" {
+		if _, exists := out[pendingName]; !exists {
+			out[pendingName] = &sdk.SourcePosition{File: relPath, Line: pendingLine}
```
bomly-guy: Same pattern as swiftpm: the in-loop path records the `version:` line (line 47) but this end-of-file flush records `pendingLine` (the package-name line), and a package block with no `version:` before the next entry is dropped mid-stream (only the last survives here). Realistically every pubspec.lock entry has a `version:`, but please align the flush with the in-loop behavior or add a comment explaining the fallback.

PR #205, opened 2026-06-27
feat: rename `--container` scan flag to `--image` (keep `--container` alias)

bomly-guy:
## What & why

`--image` is the more accurate, conventional term for scanning container images (matching `docker`, `syft`, `grype`). This renames the primary surface to `--image` everywhere a user touches it, while keeping every old "container" form working as a backwards-compatible alias — so existing scripts, CI, config files, env vars, and MCP clients keep working unchanged.

## Surfaces renamed (all four), with aliases kept

| Surface | Primary (new) | Alias kept (deprecated) |
|---|---|---|
| CLI flag | `--image` | `--container` (hidden from `--help`) |
| Env var | `BOMLY_IMAGE` | `BOMLY_CONTAINER` |
| YAML key | `target.image` | `target.container` |
| MCP arg | `image` | `container` |

## Implementation notes for reviewers

- **Deprecation notice routing.** I deliberately did **not** use cobra's `MarkDeprecated`: it prints via `OutOrStderr()`, which returns the *out* writer when one is set, so it can corrupt machine-readable `--json`/`--format` output for existing `--container` users. Instead the flag is `MarkHidden` and a one-line notice is emitted to `cmd.ErrOrStderr()` from `PersistentPreRunE` (`warnDeprecatedContainerFlag`). Verified `bomly scan --container alpine:3.20 --json` writes the warning to stderr and clean JSON to stdout.
- **Env alias.** New generic `envalias` struct tag (`internal/config/load.go`) lets `BOMLY_CONTAINER` fall back to canonical `BOMLY_IMAGE` (primary wins).
- **YAML alias.** A deprecated `target.container` leaf resolves to the same `Image` field; field order makes the canonical `target.image` win when both are present. `KnownFields(true)` still accepts both keys.
- **Concept names left intact.** `sdk.ExecutionTargetContainerImage` (value `"container-image"`, appears in JSON/SBOM output + goldens) and internal helper names are unchanged — they accurately describe the concept and avoid golden churn.
- Regenerated `docs/CONFIG_REFERENCE.md` and component docs via `make generate`.

## Tests

- Added unit tests for env-var and YAML alias resolution + precedence (`internal/config/load_test.go`).
- Kept one smoke test (`container-scan-debian`) and one root-cmd diff test on `--container` as backwards-compat guards; switched the rest to `--image`.
- `make build`, `make test` (full suite), and `make lint` (0 issues) pass. Smoke tests (`make smoke`) need network and were not run; goldens are unaffected since the flag name doesn't appear in output.

## Release

`feat:` → minor bump. New flag is additive; no breaking change (all old forms still work).

## Summary by CodeRabbit

* **New Features**
  * Container image scanning and diffing now use `--image` as the primary flag.
  * `--container` remains available as a deprecated alias for compatibility.

* **Bug Fixes**
  * Updated command validation so container image targets follow the new flag consistently.
  * Improved error messages and help examples to match current usage.

* **Documentation**
  * Refreshed getting started, scan targets, troubleshooting, and use-case examples to show `--image` throughout.


internal/mcp/server.go
```diff
@@ -118,6 +118,17 @@ func NewServer(mcpCtx Context) *server.MCPServer {
...
+func firstNonEmpty(values ...string) string {
+	for _, v := range values {
+		if v != "" {
+			return v
+		}
+	}
+	return ""
+}
+
 func jsonResult(v any) (*mcplib.CallToolResult, error) {
```
bomly-guy:
```suggestion
// firstNonEmpty returns the first non-empty string, used to prefer a primary
// argument over a deprecated alias.
func firstNonEmpty(values ...string) string {
	for _, v := range values {
		if v != "" {
			return v
		}
	}
	return ""
}

// jsonResult marshals v to JSON and returns it as a text tool result.
func jsonResult(v any) (*mcplib.CallToolResult, error) {
```

PR #216, opened 2026-06-30
Harden plugin path handling

bomly-guy:
## Summary

- Normalize and validate plugin-relative paths before joining them with install or runtime directories.
- Store downloaded plugin archives in trusted temp files instead of using remote asset names as filesystem paths.
- Add regression coverage for unsafe GitHub release asset names and manifest entrypoints.

## Why

Code scanning reported three uncontrolled path-expression findings in plugin install and checksum handling. The affected paths are derived from plugin manifests, archive metadata, or remote download metadata, so the fix moves containment checks to the trust boundary and reuses them across install, verify, test, and plugin loading paths.

## Validation

- `go test ./internal/plugin`
- `git diff --check`
- `make test`

The first full `make test` run hit transient Java readiness timeouts in Gradle, Maven, and SBT detector tests. Those packages passed on rerun, and the final `make test` passed.

### DIFF
--- a/internal/plugin/install.go
+++ b/internal/plugin/install.go
@@ -224,6 +224,7 @@ func installRemoteArchive(ctx context.Context, tempDir, source string, opts Inst
 }
 
 func safeDownloadArchiveName(name string) string {
+	// Remote archive names are only used to preserve a trusted extension for temp-file format detection.
 	name = strings.ReplaceAll(strings.TrimSpace(name), "\\", "/")
 	if name == "" {
 		return ""
@@ -421,6 +422,7 @@ func extractArchiveEntry(name, targetDir string, mode os.FileMode, write func(st
 	if err := os.MkdirAll(filepath.Dir(destination), 0o755); err != nil {
 		return fmt.Errorf("create plugin archive parent for %q: %w", name, err)
 	}
+	// Symlinks would make the lexical containment checks above vulnerable to writes outside targetDir.
 	if mode&os.ModeSymlink != 0 {
 		return fmt.Errorf("plugin archive entry %q uses unsupported symlink mode", name)
 	}
--- a/internal/plugin/install.go
+++ b/internal/plugin/install.go
@@ -147,7 +147,10 @@ func installDevBinary(ctx context.Context, tempDir, source string) (Manifest, st
 		return Manifest{}, "", err
 	}
 	entry, _ := entrypointForManifest(manifest)
-	targetBinary := filepath.Join(tempDir, entry)
+	targetBinary, err := pathInPluginDir(tempDir, entry)
+	if err != nil {
+		return Manifest{}, "", fmt.Errorf("plugin entrypoint %q must stay within the plugin directory", entry)
+	}
 	if err := os.MkdirAll(filepath.Dir(targetBinary), 0o755); err != nil {
 		return Manifest{}, "", fmt.Errorf("create plugin binary dir: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


bomly-guy (commented):
Reviewed via /code-review skill. Approve overall — solid, well-targeted fix for the path-traversal findings, reused consistently across install/verify/test/load paths, with good regression test coverage. Left a few inline suggestions, none blocking.

internal/plugin/install.go
```diff
@@ -366,19 +404,18 @@ func extractTarGzArchive(archivePath, targetDir string) error {
 }
 
 func extractArchiveEntry(name, targetDir string, mode os.FileMode, write func(string) error) error {
-	cleanName := filepath.Clean(filepath.FromSlash(strings.TrimSpace(name)))
-	if cleanName == "." || cleanName == "" {
+	if strings.TrimSpace(name) == "" || strings.TrimSpace(name) == "." {
 		return nil
 	}
-	if filepath.IsAbs(cleanName) || strings.HasPrefix(cleanName, "..") {
+	destination, err := pathInPluginDir(targetDir, name)
```
bomly-guy, line 411: `pathInPluginDir`'s containment check is purely lexical (cleans the entry path, then checks with `filepath.Rel`). On its own that would leave a TOCTOU gap: a malicious archive could plant a symlink entry and then a later entry that writes through it, escaping `targetDir` despite the lexical check passing. I see `extractArchiveEntry` rejects `mode&os.ModeSymlink != 0` further down in this function, which closes that gap — nice catch. Worth a comment on that symlink-rejection line making the intent explicit (e.g. "reject symlinks to preserve the extraction containment guarantee"), since it's easy for a future contributor to read it as an arbitrary unsupported-type restriction and loosen it without realizing it's load-bearing for path safety.

internal/plugin/types.go
```diff
@@ -421,15 +453,25 @@ func validateManifest(manifest Manifest) error {
...
-	if filepath.IsAbs(cleanEntry) || strings.HasPrefix(cleanEntry, "..") {
+	if _, err := pathInPluginDir(".", entry); err != nil {
 		return fmt.Errorf("plugin entrypoint %q must stay within the plugin directory", entry)
 	}
 	return nil
 }
 
 func checksumFile(path string) (string, error) {
-	file, err := os.Open(path)
+	cleanPath, err := filepath.Abs(path)
```
bomly-guy, line 464: `filepath.Abs` + `os.Stat`/`IsDir` is a useful correctness improvement but isn't actually a containment check — it just resolves and confirms a regular file. That's fine given current callers only ever pass already-validated/temp paths, but the function name might invite reuse later on unvalidated input. A one-line comment noting `checksumFile` assumes the caller already constrained `path` would prevent future misuse.

internal/plugin/install.go
```diff
@@ -218,6 +223,32 @@ func installRemoteArchive(ctx context.Context, tempDir, source string, opts Inst
 	return installArchiveAtPath(ctx, tempDir, archivePath, source, opts.Checksum, opts.InsecureSkipChecksum)
 }
 
+func safeDownloadArchiveName(name string) string {
```
bomly-guy, line 226: Nice belt-and-suspenders sanitization, but worth noting: since the downloaded archive is now always written via `os.CreateTemp` rather than joined with the remote-controlled name, `downloadName` is only ever used to pick a file extension here — this function no longer gates an actual filesystem path. A short comment clarifying its remaining job is "extension sniffing only" would help a future refactor avoid assuming it's still load-bearing for path safety.

PR #246, opened 2026-07-08
feat: revamp MCP server with compact agent-optimized responses

### LINKED ISSUE #245: MCP bomly_scan response is too large for agent tool-result limits (verbosity + duplicated package objects)
## Summary

The `bomly_scan` MCP tool returns the full `bomly scan --format json` payload verbatim. For a realistic project this is large enough to **exceed an MCP client's tool-result size limit and get truncated to an error**, so the agent's scan call returns nothing usable and it has to retry — burning tokens and turns for no result.

Measured on a 135-package npm project (one app, `--enrich --audit`):

```
total scan JSON: ~1.8 MB  (2.9 MB for a larger lockfile)
  findings     1,605,519 bytes  (89%)
  packages       130,855 bytes  (7%)
  manifests       60,008 bytes  (3%)
```

Observed directly in a real agent session (Claude Code): 4 of 8 `bomly_scan` calls came back as
`Error: result (1,714,343 characters ...) exceeds maximum allowed tokens. Output has been saved to <file>`
— i.e. the tool result was dropped. Only the smaller re-scans of individual sub-paths returned usable JSON.

## Where the size comes from

1. **`findings[]` embeds a full `package` object per finding (~1.9 KB each, 50 findings), duplicating the top-level `packages[]` array.** Each finding carries `package`, `references`, `cvss`, `epss`, `cwes`, `data_source`, `namespace`, full `description`, etc. — most of which an agent doing remediation doesn't need to see inline.
2. **Every scanned package is listed, clean or not** (135 packages, only 15 vulnerable), each with license + location metadata.
3. Advisory reference URL lists and descriptions are carried in multiple sections.

## Impact

This is exactly the workload the MCP server is meant to serve — an agent scanning a project to remediate vulnerable deps. At its current size the primary scan call is unreliable over MCP, which makes the tool look worse than it is (the agent gets an error, not an answer, on its most important call).

## Suggested fixes (in impact order)

1. **A compact, MCP-appropriate `bomly_scan` response**: vulnerable packages only; per vuln the advisory IDs (id + aliases), `fixed_versions`/`min_safe_version`, `fix_state`, severity, and the dependency path — and drop clean packages, license/location data, `epss`/`cwes`/`cvss` vectors, and full reference-URL lists (or make them opt-in). `bomly_vuln_fix_context` already returns roughly this shape and is ~4 KB — that's the right size envelope for an agent tool result.
2. **Stop embedding a full `package` object inside every `findings[]` entry** — reference it by purl/name+version instead. This de-dupes findings against the top-level `packages[]` and helps CLI/JSON consumers too, not just MCP.
3. Consider a `--summary`/severity-threshold mode on the scan JSON so callers can ask for just what crossed the audit threshold.

## Note on scope / study context

Found while building the CS-A agent study (bomly-dev/bomly-agent-study), which measures agents fixing vulnerable deps with vs. without the bomly MCP server. This isn't a study blocker — the agents still complete via sub-path re-scans — but it costs real tokens/turns and depresses the MCP condition's efficiency numbers. We are NOT asking for recursive/root discovery here (that's tracked separately as E1); the study harness already scans each fixture path separately. This issue is purely about the size/shape of a single scan's response.

bomly-guy:
Closes #245.

## Problem

`bomly_scan` over MCP returned the full `scan --format json` document verbatim — ~1.8MB on a realistic enriched npm project — exceeding MCP client tool-result limits and truncating the agent's primary call to an error. The issue comment adds real-session usability gaps: scoped names collapsed (`@tailwindcss/postcss` → `postcss`), no actionable-vs-informational separation, weak transitive fix advice, no branch-aware delta, opaque "no subprojects discovered" errors.

## What changed

**Commit 1 — CLI JSON aligned with the SDK three-collection model** (`refactor(output)`)
- `findings[]` are now pure references, mirroring `sdk.Finding`: identity-only `package` ref (join `packages[]` by purl), `vulnerability_id`, `dependency_refs`. All flat advisory copies (description/CVSS/EPSS/references/fix data) dropped — that data lives once in `packages[]`. No compat shims.
- Ecosystem-native `DisplayName()` on `sdk.Coordinates` (`@org/name` for npm, `org/name` for Go/Composer, `org:name` otherwise); scoped names are preserved everywhere. `ScanPackageEntry` gains `org`.
- `DiffResponse` gains a `packages[]` collection (base∪head, head wins) so diff findings join like scan findings. Renderers (text/markdown/TUI) converted to registry joins, the way SARIF always worked.
- Measured on a 70-package npm fixture with 16 findings: full JSON is 151KB (findings no longer dominate).

**Commit 2 — compact `bomly_scan`** (`feat(mcp)`)
- New agent-facing response (`schema_version "mcp/1"`): findings classified (`fix_available` / `no_fix_upstream` / `wont_fix` / `policy_only`), grouped into ranked remediation groups — one concrete change (which direct dep, which manifest, which version) and every finding it closes. Ranked KEV → severity → EPSS → fixability. Shortest paths via bounded upward BFS. Hard caps with explicit truncation counters; a size test pins a 15-vuln scan under 8KB.
- Pipeline warnings + manifest resolution fallbacks surface as `diagnostics`; "no subprojects discovered" errors (CLI too) now include a discovery probe naming the manifest evidence that exists and which package manager it belongs to.

**Commit 3 — integrated fix context; compact diff & explain** (`feat(mcp)`)
- `bomly_vuln_fix_context` removed; its context is integrated everywhere. Remediation groups carry fix kinds (`direct-bump` / `transitive-override` / `lockfile-refresh` / `no-fix-upstream` / `policy-review`) and per-package-manager override advice (npm/pnpm/yarn/maven/gradle/gomod/cargo/pip/poetry/bundler/composer).
- `bomly_diff` returns the branch-aware security delta (introduced / resolved / persisted, keyed by advisory id independent of version bumps) plus remediation for everything still open after merge.
- `bomly_explain` is the bounded drill-down: full advisory detail (descriptions, references, CVSS, ranges) for the queried package only, plus paths and remediation.
- Docs: `docs/MCP.md` rewritten (4-tool surface, compact contract), `docs/SCHEMAS.md`, decision-log entries in `dev-docs/ARCHITECTURE.md`, contributor checklists.

## Verification

- `make test`, `make lint`, `make generate` green.
- End-to-end against a vulnerable npm fixture via a real stdio MCP client: compact scan = **9.2KB** (full document 151KB), remediation groups with correct manifest/version targets and npm override advice, explain returns full advisory detail for one package (16KB), empty-dir scan returns the discovery probe, 4 tools listed.
- Smoke goldens regenerated (`make smoke ARGS="-update"`); **`scan-gradle.golden.json` needs regeneration in CI** — this machine has no Java runtime.

## Notes for reviewers

- Full scan JSON is intentionally unreachable over MCP; the CLI is the artifact channel (`bomly scan --format json -o <file>`), stated in tool descriptions and response hints.
- No schema version bump per maintainer decision (no external consumers yet); `mcp/1` tags the MCP shapes independently.
- Follow-ups tracked separately: CI-readiness hints (pnpm minimumReleaseAge-style gates) and per-detector "skipped because" instrumentation.

## Summary by CodeRabbit

* **New Features**
  * MCP scan, explain, and diff outputs are now compact, drill-down friendly, and use a clearer truncation model for large results.
  * Package names display in ecosystem-native formats across reports and the UI for improved readability.
* **Bug Fixes**
  * Advisory, reachability, and fix details now resolve more consistently across scan/diff/explain views.
  * Diff counters and license-finding tallies now deduplicate and group more accurately.
* **Documentation**
  * Updated MCP and schema docs to match the new compact response structure and findings projections.
* **Chores**
  * Streamlined the MCP toolset by removing the vulnerability fix-context tool.


internal/cli/mcp_cmd.go
```diff
@@ -335,7 +390,7 @@ func (a *mcpOptionsAdapter) RunDiff(ctx context.Context, req mcp.DiffRequest) (o
 
 	baseTarget, headTarget, projectIdentifier, _, _, err := resolveGitDiffGraphs(ctx, o, nil, logger, req.Base, req.Head)
```
bomly-guy: This still forces every `bomly_diff` MCP request through the Git resolver even when the caller supplied `image`/`container`. The tool schema and docs now advertise container diffs (`image` with base/head tags or digests), but this path calls `resolveGitDiffGraphs`, so a request like `{image:"alpine", base:"3.19", head:"3.20"}` tries to verify `3.19` and `3.20` as refs in the local repo instead of resolving `alpine:3.19`/`alpine:3.20`. The CLI command already switches to `resolveContainerDiffGraphs` when `current.Image != ""`; the MCP adapter needs the same branch or should stop exposing the image argument.

bomly-guy (commented):
Found one functional issue worth fixing before merge.

internal/cli/mcp_cmd.go
bomly-guy: Good catch — fixed in 48a07e2. `RunDiff` now branches on the resolved image target and calls `resolveContainerDiffGraphs` when `image`/`container` is set, mirroring the CLI diff command's `current.Image != ""` case.

Verified end-to-end over MCP: `{image:"alpine", base:"3.19", head:"3.20"}` now resolves the container tags and returns the package delta (14 changed, 1 removed) instead of failing to find Git refs `3.19`/`3.20` in the local repo.

PR #254, opened 2026-07-13
feat(scan): recursive manifest discovery with --recursive/--max-depth/--exclude

bomly-guy:
Implements E1 — recursive manifest discovery. Until now, subproject discovery inspected only the scan root (`planFilesystemSubprojects`), so real monorepo support depended entirely on package managers' own workspace expansion.

## What's new

- **`--recursive`** (scan / explain / diff, default off): walks the directory tree below the scan root and plans a subproject for every directory with recognized manifest evidence, consolidated into one graph.
- **`--max-depth N`**: bounds the walk. Root = depth 0, direct child = 1; default **3**; `0` = unlimited. Requires `--recursive` (explicit flag without it → exit 4).
- **`--exclude <glob>`** (repeatable / CSV): skips subtrees on top of built-in ignore rules (dot-dirs, `node_modules`, `vendor`, `target`, `build`, `dist`, `__pycache__`, `pyvenv.cfg` virtualenvs). Patterns with `/` match the root-relative path, without `/` match basenames at any depth; `path.Match` syntax, no `**`.
- **Per-PM ancestor pruning**: workspace/reactor-expanding managers (maven, gradle, npm, pnpm, yarn, cargo, sbt, mix) are not re-planned below an ancestor that already covers them — their nested modules resolve through the root, so no double-counting. `gomod` and non-workspace managers never prune (a nested `go.mod` is an independent module). Pruning never hides other ecosystems' nested manifests.
- **Consolidation fix**: manifest paths are now rebased onto the subproject's `RelativePath` for core detectors (mirroring the pre-seeded `rebaseGraphLocations`). This fixes a latent dedup-key collision where two subprojects each emitting `requirements.txt` collapsed to one key and silently dropped a manifest.
- **Better exit-5 hints**: non-recursive runs that find nested evidence now say `hint: manifests exist in subdirectories (e.g. fixtures/webapp); retry with --recursive`; recursive runs describe their search bounds. The diagnostic probe shares the walk's skip rules so hints and discovery never disagree.
- **Observability**: INFO walk start/summary (dirs visited/skipped, subprojects, pruned, duration) at `-v`; per-directory exclude/prune/plan decisions at `-vv`; unreadable dirs WARN and never abort.

## Config surface

[...]

PR #256, opened 2026-07-13
test(smoke): recursive discovery smoke against bomly-agent-study

bomly-guy:
Stacked on #255 (which stacks on #254). End-to-end smoke coverage for recursive manifest discovery, pinned to `bomly-dev/bomly-agent-study@fc32147b` (tag `v1-fixtures-final`) — a real repo with no root manifest and independent projects in nested directories.

- **`scan-recursive-monorepo`** (golden): `--recursive --exclude harness,fixtures/service` resolves `fixtures/api-java` (maven native, 16 deps) and `fixtures/webapp` (npm lockfile, 135 deps) as distinct subprojects with repo-relative manifest paths. The pip projects are deliberately excluded: they resolve through a live `pip install` (no committed `requirements.lock` fast-path), which is not environment-deterministic; the exclusion doubles as `--exclude` coverage. Same-named-manifest dedup across nested subprojects is covered by consolidation unit tests in #254.
- **`scan-recursive-depth-exclude`** (golden): `--max-depth 2` + excludes down to just the npm subproject; runs without any package-manager binaries.
- **`TestScanRecursiveNothingAtRootWithoutFlag`** (no golden): the same repo without `--recursive` exits 5 and stderr carries the new `retry with --recursive` hint — locking in the root-only default.

Goldens regenerated with `-update` and verified stable across two replay runs; they contain no temp-clone paths (manifest paths are repo-relative post-rebase, so `normalizeManifestPaths` needed no changes).


bomly-guy:
Per review: replaced the `scan-recursive-depth-exclude` golden case with `TestScanRecursiveDepthLimitFindsNothing` (2385fa1) — `--max-depth 1 --exclude harness` leaves nothing discoverable, so it asserts exit 5 and the bounded-search stderr message instead of comparing a golden. Default-depth discovery of nested projects stays covered by `scan-recursive-monorepo`.

PR #270, opened 2026-07-14
feat(scan): subproject & module presentation with per-module manifests

bomly-guy:
With recursive discovery shipped (#254–#256), subprojects genuinely represent nested paths — but views hadn't caught up: the interactive components tab showed one flat "project" node, and multi-module projects were worse, since their detectors merged everything into a single root manifest with module structure invisible everywhere. This PR makes the whole surface reflect real project layout, delivered as an ordered commit sequence (one reviewable step each).

## Terminology

- **subproject** — an independently discovered nested directory (what `--recursive` finds; `Subproject.RelativePath != "."`).
- **module** — a member the package manager natively resolves under one root manifest (Maven reactor module, npm/pnpm workspace member, Cargo workspace member).

## Commits

**1. `feat(output)` — hierarchy derivation + counting fixes.** New `output.ClassifyManifest`/`BuildHierarchy`: the project → subproject → module grouping is derived purely from each manifest's existing `subproject` + `path` fields — **no JSON schema changes**; consumers can apply the same rule. Rides along: the text report's "in M manifests" now counts manifests instead of graph roots, the misleading `Discovered 1 subproject: . (npm)` banner on plain scans is gone (replaced by a grouped manifest tree, rendered only when a scan actually spans subprojects/modules), and markdown package totals + inventory dedup shared packages by PURL/ID instead of double-counting. MCP compact summary gains additive `subprojects`/`modules` counts (`mcp/1` unchanged, `omitempty`).

**2. `feat(tui)` — subproject/module nodes.** Components tree and manifest list group manifests under subproject/module nodes (string-keyed, default-expanded, collapsible); the glyph helper is ancestors-aware instead of hardcoding two levels; flat single-root scans keep exactly the previous tree. Source tab gains a subprojects section and per-manifest subproject/module attributes; group nodes get their own details panes; the diff components tab group-by cycle gains a **subproject** axis.

**3. `feat(npm,pnpm)` — one manifest per workspace member.** Shared `detectors.SubgraphFrom` partitions a merged graph by module root. npm's v2/v3 packages-map parser is now workspace-aware (member dirs → application roots; `link:` entries resolve to member nodes instead of synthesizing versionless ghosts; member devDependencies produce scoped edges — all previously dropped or mangled). pnpm wires non-`"."` importers (previously **silently dropped**) with `link:` workspace deps resolving to member roots.

[...]

PR #271, opened 2026-07-16
feat(gradle): per-module manifests for multi-project builds + first-party enrichment skip

### LINKED ISSUE #253: First-party packages: skip external enrichment; expand Gradle multi-project modules
## Context

Follow-up from #252, which fixed the Maven TGF parser to handle multi-module reactors (`mvn dependency:tree -DoutputType=tgf` emits one concatenated block per module). That PR is intentionally scoped to the parser; this issue tracks a separate modeling decision it surfaced.

This is **not a bug** — nothing is broken and the graph shape is already correct. It's a question of how faithfully we represent a reactor's own modules in the data model.

## Current behavior

Each reactor module's own artifact (e.g. `com.bomly:module-a@1.0.0`) is emitted by Maven as the scopeless root of its TGF block. The parser records it as an ordinary `package`-typed graph node with a valid Maven PURL. Consequences:

- **Tree shape is fine.** The module artifacts are graph roots (no incoming edges), and consolidation already wraps multiple roots under a synthetic `manifest` virtual root (`internal/engine/consolidation/manifest.go`).
- **They are enriched.** Matchers iterate `Graph.Nodes()` and query OSV / deps.dev / ClearlyDefined for every node with a non-empty PURL (`internal/matchers/osv/matcher.go`). A reactor module is a first-party artifact absent from public advisory/registry data — so these are wasted lookups, with a small risk of a coincidental false name match.
- **They are counted** in the package totals.
- **They appear in diffs.** Bumping your own module version reads as a dependency change (`NodeIsDiffable`, `sdk/graph.go`, does not currently exclude them because they're typed `package`).

Note this predates multi-module support: a single-module Maven scan already lists the project artifact (e.g. `com.bomly:example-java-maven`) in its `packages` output. Multi-module simply multiplies it by the module count.

## Relevant model mechanics

- `NodeIsDiffable` (`sdk/graph.go`) excludes `manifest` and `application` typed nodes from diffs.
- `Coordinates.CanonicalPURL()` (`sdk/purl.go`) suppresses the PURL **only** for `manifest` — so `manifest` nodes are skipped by matchers, `application` nodes are not.
- Consolidation expects **at most one** `application` root per graph; multiple `application` roots are flagged anomalous by `logUnexpectedMultiRootGraph` (`internal/engine/pipeline.go`). So blanket-typing every module `application` is not viable.

## Options considered

1. **Type each module artifact as `manifest`.** Smallest change that removes the noise: suppresses their PURL (matchers skip them), excludes them from diffs, and stays compatible with the virtual-root path (doesn't trip the single-application-root expectation). Trade-off: the modules also drop out of the `packages` list and generated **SBOMs**, where a first-party module is often a component you *want* described.

2. **Split the reactor into one manifest / `GraphEntry` per module.** Most faithful to representing modules "as manifests" — each submodule becomes a first-class subproject with its own root, mirroring subproject discovery. Larger detector change (split concatenated blocks into per-module graphs, synthesize per-module manifest metadata/paths) and the biggest golden churn.

3. **Leave as-is.** Accept the enrichment/diff/count noise as the cost of keeping module artifacts visible as packages/SBOM components.

## Deciding factor

Whether reactor modules should be visible to SBOM consumers as first-party components (argues against suppressing their PURL — favors option 2 or 3) or treated as internal scaffolding to hide from vuln scanning and diffs (favors option 1).

## Notes

- Any typing change regenerates the `scan-maven` and `scan-java-maven-reachability` smoke goldens, which requires Java on the runner (CI, not local).
- The same reasoning likely applies to Gradle multi-project builds.

bomly-guy:
Closes #253 (as re-scoped).

## 1. Skip external enrichment for first-party packages — `fix(match)`

`application`-typed graph nodes (workspace members, reactor modules, the project's own package) are absent from public advisory/registry sources, yet matchers still queried them — wasted lookups plus a coincidental false-name-match risk (a member named like a real npm package would adopt its advisories).

- New `sdk.NodeIsEnrichable` predicate (mirrors `NodeIsDiffable`): false for `manifest` + `application` types.
- Applied at every enrichment path: `matchers.RegistryPackagesForGraph` (deps.dev, scorecard, builtin grype), the OSV matcher's graph iteration, and external grype's result mapping.
- External grype's SBOM **input** is deliberately unfiltered — `sbom.FromDepGraph` is shared with user-facing SBOM generation, where first-party components must remain — so first-party matches are dropped when results map back in.
- First-party entries stay in the `packages` inventory and SBOMs, just unenriched. External plugin matchers (ClearlyDefined, EOL) can adopt the predicate in their repos.
- **Zero golden churn**, verified by re-running the enrich/audit/reachability smokes (`scan-go-enrich`, `scan-go-audit(-high)`, `scan-{go,npm,java-maven}-reachability`) plus the workspace cases against unchanged goldens.

## 2. Gradle multi-project module support — `feat(gradle)` (reverses the #270 deferral)

The gradle detector ran the root `dependencies` task only, so subproject trees were invisible — while recursive discovery pruned nested gradle modules on the assumption the root detector expands them: multi-project builds silently under-reported.

- `walkGradleSettingsModules`: regex walk of `settings.gradle(.kts)` `include(...)` (both forms), `projectDir` overrides honored, escaping paths skipped, composite `includeBuild` out of scope.
- **One** invocation covers every subproject: `gradle dependencies :app:dependencies :lib:dependencies --console=plain`. `Root project 'x'` / `Project ':x'` banners switch graph attribution.
- `project :x` tokens — including the colon-less `project x (n)` form declared-only listings print — resolve to the subproject's synthesized `application`-typed root node: real inter-module edges (mirrors maven web→core), no placeholder leaves.
- Per-subproject `GraphEntry` split (`SubgraphFrom`), manifest paths `<dir>/build.gradle(.kts)` — modules flow into the #270 module presentation everywhere (TUI, text tree, hierarchy, counts) with no view changes.
- Layered degrade: settings-walk failure → root-only resolve; failed multi-task run (stale settings) → root-only retry; unseen subprojects add no orphan nodes; single-project builds keep one entry (regression-pinned).
- Verified against a real Gradle 8.7 build: 3 manifests (`build.gradle`, `app/build.gradle`, `lib/build.gradle`), full transitive closures, counts hold (11 = 4 direct + 7 transitive = 9 runtime + 2 dev); `--recursive` on a repo nesting the build yields one subproject with rebased per-module manifests.

## 3. Smoke + docs

- New `scan-gradle-multimodule` smoke case pinned to `bomly-dev/example-java-gradle-multimodule` @ v1.0.0 (repo commits its gradle wrapper; only `java` required). **The fixture repo + golden are pending** — repo creation needs a human (scaffold is ready locally); the golden lands in a follow-up commit on this branch once the repo exists. CI smoke is schedule/dispatch-only, so PR checks are unaffected.
- Decision-log entries for both changes; SCAN_TARGETS/MODELS updates; CLAUDE/AGENTS matcher-list correction (ClearlyDefined/eol are external plugins).

## Summary by CodeRabbit

* **New Features**
  * Gradle multi-project scans now emit separate module-level dependency manifests, with cross-module relationships preserved.
* **Bug Fixes**
  * External vulnerability enrichment (OSV/Grype) now correctly skips first-party/workspace artifacts, improving match accuracy and preventing first-party identifiers from being enriched.
  * Scoped Gradle dependency parsing was improved to keep expected runtime module relationships while avoiding unwanted targets.
* **Documentation**
  * Updated architecture, models, and scan-target docs to reflect Gradle module discovery/manifest behavior and first-party handling in output and SBOMs.


bomly-guy:
Fixture repo [example-java-gradle-multimodule](https://github.com/bomly-dev/example-java-gradle-multimodule) is now populated and tagged `v1.0.0` (authored as ahmed@bomly.dev per fixture convention). Pushed de0126c with the generated `scan-gradle-multimodule` golden: three per-module manifests (`build.gradle`, `app/build.gradle`, `lib/build.gradle`), 14 packages, shape matching the merged maven-multimodule golden. Verified locally with no drift in scan-npm-reachability (enrichment-skip path) or scan-maven-multimodule.

internal/detectors/gradle/detector.go
```diff
@@ -279,45 +387,146 @@ func depGraphFromGradleOutput(raw []byte, rootName string) (*sdk.Graph, error) {
...
 			continue
 		}
+		// `project :x` tokens (colon-less in declared-only listings) become
+		// edges to the subproject's own root node when the settings walk
+		// knows the project, so inter-module dependencies stay real edges
+		// instead of placeholder leaves.
+		if projectPath, isProject := gradleProjectRef(token); isProject {
+			if moduleRootID, err := roots.ensure(projectPath); err != nil {
+				return gradleParseResult{}, err
+			} else if moduleRootID != "" {
```
bomly-guy: **[P1] Preserve the scope on inter-project dependencies.** `roots.ensure` returns an application node with no scope, and this branch adds only the edge. The engine subsequently calls `FilterDetectionResultByScope`; in an `app -> project :lib` runtime graph, `:lib` is neither the app entry's root nor `ScopeRuntime`, so it and its whole subtree are removed. I reproduced this by parsing a runtime report containing `project :lib` and applying `sdk.FilterDetectionResultByScope(..., sdk.ScopeRuntime)`. Please carry `currentScope` onto the project reference (with project-local node metadata, so scopes do not leak between entries) before scope filtering.

sdk/graph.go
```diff
@@ -466,6 +466,25 @@ func NodeIsDiffable(node *Dependency) bool {
...
+// published to public sources, so querying them wastes lookups and risks
+// coincidental name matches; they remain in the packages inventory and in
+// generated SBOMs, just without external enrichment. External plugin matchers
+// should apply the same predicate to the nodes they iterate.
+func NodeIsEnrichable(node *Dependency) bool {
+	if node == nil {
+		return false
+	}
+	switch node.Type {
+	case PackageTypeManifest, PackageTypeApplication:
```
bomly-guy: **[P1] Do not use component type as an ownership signal.** SBOM imports preserve CycloneDX/SPDX component types in `internal/sbom.ToGraph`, and `application` there is an artifact kind, not proof that the component belongs to the scanned workspace. A third-party application component with a PURL will now be skipped by OSV, deps.dev, scorecard, and Grype, silently suppressing vulnerability data. Please base first-party suppression on detector/root provenance (or another explicit ownership marker) rather than excluding every `PackageTypeApplication` node.

bomly-guy (commented):
Two correctness issues found in the current head.

sdk/graph.go
bomly-guy: Fixed in 646d8ea. Ownership is now an explicit `Coordinates.FirstParty` marker set only by detectors at the sites that synthesize or match the build's own nodes (root packages, workspace members, reactor modules, gradle subproject roots, plus the diff audit synthetic root). `NodeIsEnrichable` keys on the marker and manifest typing only — `sbom.ToGraph` never sets it, so an application-typed component imported from a CycloneDX/SPDX document keeps flowing to OSV, deps.dev, scorecard, and Grype. Regression test covers the ToGraph → RegistryPackagesForGraph path (`TestRegistryPackagesForGraphKeepsImportedApplicationComponents`). Note `NodeIsDiffable` still excludes application-typed nodes from diffs — that predates this PR; happy to migrate it to the marker in a follow-up if you want imported application components diffable.

internal/detectors/gradle/detector.go
bomly-guy: Fixed in f99efb8. The parser now builds one graph per project section, and `project :x` tokens become project-local reference nodes carrying the section's scope — in your repro, the `project :lib` edge inside :app's runtimeClasspath is ScopeRuntime, so `sdk.FilterDetectionResultByScope(..., ScopeRuntime)` keeps :lib and its subtree. Because the reference is a fresh instance (sharing only the module root's coordinates/ID), the scope cannot leak into :lib's own entry; consolidation still merges the two faces into one package. Pinned by `TestGradleMultiProjectRuntimeScopeFilterKeepsProjectEdges` (your exact reproduction over `subprojectGraphEntries`) and instance-isolation assertions in the multi-project fixture test.

internal/detectors/gradle/detector.go
```diff
@@ -102,19 +109,61 @@ func (d Detector) ResolveGraph(ctx context.Context, req sdk.DetectionRequest) (s
...
+	for _, moduleRoot := range parsed.modules {
+		moduleGraph, err := detectors.SubgraphFrom(parsed.graph, moduleRoot.rootID)
+		if err != nil {
+			return nil, fmt.Errorf("extract subgraph for %s: %w", moduleRoot.module.ProjectPath, err)
+		}
+		AttachGradlePositions(moduleGraph, filepath.Join(workingDir, filepath.FromSlash(moduleRoot.module.Dir)))
+		entries = append(entries, sdk.GraphEntry{
+			Graph:    moduleGraph,
+			Manifest: sdk.ManifestMetadata{Path: moduleRoot.module.Dir + "/" + moduleRoot.module.ManifestFile, Kind: sdk.ManifestKind(moduleRoot.module.ManifestFile)},
+		})
```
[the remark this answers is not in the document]
bomly-guy: Fixed in f99efb8 — the parser now builds one graph per project section with project-local node instances (no shared pointers, no SubgraphFrom), so a package that is runtime in :app and test-only in :lib keeps exactly one scope per entry, and AttachGradlePositions runs per project graph so file positions cannot leak between manifests. The regenerated golden shows the previously-leaked root build.gradle location gone from module entries. Instance isolation is asserted directly in the multi-project fixture test.

internal/detectors/gradle/detector.go
```diff
@@ -143,10 +201,33 @@ func (d Detector) resolveGraph(ctx context.Context, stderr io.Writer, projectPat
...
+	}
+	// A settings parse can name subprojects the build no longer has (or a
+	// composite/scripted layout the regex walk misread), failing the whole
+	// multi-task invocation. Degrade to the root-only report.
+	logger.Warn("gradle multi-project dependency report failed; retrying the root project only", zap.Error(err))
+	executable, args, cmdErr := d.commandSpec(workingDir, nil)
+	if cmdErr != nil {
+		return gradleParseResult{}, fmt.Errorf("resolve gradle command: %w", cmdErr)
+	}
+	return d.runDependencies(ctx, stderr, workingDir, verbose, executable, args, nil)
```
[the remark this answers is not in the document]
bomly-guy: Fixed in f99efb8. Both the primary multi-project invocation and the root-only fallback now go through a shared `runScopedThenFull` ladder, so a --scope runtime scan retries the fallback with `--configuration runtimeClasspath` first and never silently widens. Pinned by `TestResolveGraphMultiTaskFailureFallbackKeepsScope`, which asserts the recorded fallback args.

internal/detectors/gradle/modules.go
```diff
@@ -0,0 +1,169 @@
...
+	Group        string
+	ManifestFile string
+}
+
+var (
+	gradleParenthesizedInclude = regexp.MustCompile(`(?s)\binclude\s*\((.*?)\)`)
+	gradleLineInclude          = regexp.MustCompile(`(?m)^\s*include\s+([^\r\n]+)$`)
+	gradleQuotedValue          = regexp.MustCompile(`["']([^"']+)["']`)
+	gradleProjectDirOverride   = regexp.MustCompile(`(?m)project\s*\(\s*["']([^"']+)["']\s*\)\.projectDir\s*=\s*file\s*\(\s*["']([^"']+)["']\s*\)`)
+	gradleGroupAssignment      = regexp.MustCompile(`(?m)^\s*group\s*=\s*["']([^"']+)["']`)
```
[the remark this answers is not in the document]
bomly-guy: Fixed in 84d5779. Declaration extraction now runs over `stripGradleComments` output — a lightweight string-aware scanner removing // and /* */ comments while preserving string literals and line structure — and the include/projectDir regexes are line-anchored so include-shaped text inside unrelated string literals can't register modules either. Fixtures cover line comments, block comments, trailing comments, //include, and a `println "include(':fake')"` string case (`TestWalkGradleSettingsModulesIgnoresCommentedIncludes`, `TestStripGradleComments`).

internal/detectors/gradle/modules.go
```diff
@@ -0,0 +1,169 @@
...
+
+// walkGradleSettingsModules reads the settings script in workingDir and
+// returns every included subproject in project-path order. A build without a
+// settings script (or with no includes) yields no modules and no error;
+// composite builds (includeBuild) are not expanded. Included paths that
+// resolve outside the build root are skipped.
+func walkGradleSettingsModules(workingDir string) ([]gradleModule, error) {
+	body, err := readGradleSettings(workingDir)
+	if err != nil || body == "" {
+		return nil, err
```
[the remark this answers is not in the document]
bomly-guy: Fixed in f99efb8 + 84d5779: `walk gradle settings modules: %w` on the settings read, `map gradle dependency output: %w` in runDependencies, and `register subproject graph for banner/project reference %q: %w` at the two roots.ensure call sites (which themselves wrap with `add subproject root %q: %w`). All remain unwrappable.

internal/detectors/gradle/detector_test.go
```diff
@@ -281,18 +282,154 @@ func TestRunDependencies_UsesSettingsGradleRootName(t *testing.T) {
...
+	writeGradleFile(t, projectDir, "lib/build.gradle", "dependencies {}\n")
+
+	fixture, err := os.ReadFile(filepath.Join("testdata", "dependencies-multiproject.txt"))
+	if err != nil {
+		t.Fatalf("read fixture: %v", err)
+	}
+	argsFile := filepath.Join(projectDir, "gradle-args.txt")
+	script := "#!/bin/sh\necho \"$@\" > " + argsFile + "\ncat <<'FIXTURE_EOF'\n" + string(fixture) + "FIXTURE_EOF\n"
+	if err := os.WriteFile(filepath.Join(projectDir, "gradlew"), []byte(script), 0o755); err != nil {
+		t.Fatalf("write gradlew fixture: %v", err)
```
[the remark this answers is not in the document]
bomly-guy: Fixed in 1a5cc59. The gradle package now builds its fake gradle and java executables once in TestMain (mirroring internal/cli/root_cmd_test.go), parameterized per test via BOMLY_FAKE_GRADLE_{FIXTURE,ARGS_FILE,FAIL_ON} / BOMLY_FAKE_JAVA_FAIL env vars. Argument capture and fixture behavior are preserved, and since the fakes are Go binaries the previous unix-only skips on the ResolveGraph tests are gone. (Used a local builder mirroring the cli package's buildHelperBinary — testutil.BuildGoBinary needs a testing.TB, which TestMain doesn't have.)

docs/SCAN_TARGETS.md
```diff
@@ -84,7 +84,7 @@ Scan output distinguishes two kinds of nesting:
 - A **subproject** is an independently discovered nested directory (its own detector run) — what `--recursive` finds.
 - A **module** is a member the package manager natively resolves under one root manifest: a Maven reactor module, an npm/pnpm workspace member, a Cargo workspace member.
 
-The npm, pnpm, cargo, and maven detectors emit **one manifest entry per module** — `apps/web/package.json`, `crates/api/Cargo.toml`, `core/pom.xml` — alongside the root manifest, each carrying the module's reachable dependency subtree (a virtual Cargo workspace root emits member entries only). Detectors without per-module emission (gradle, sbt, mix, yarn classic, pub, and the node *native* detectors) keep one merged root manifest.
+The npm, pnpm, cargo, maven, and gradle detectors emit **one manifest entry per module** — `apps/web/package.json`, `crates/api/Cargo.toml`, `core/pom.xml`, `app/build.gradle` — alongside the root manifest, each carrying the module's reachable dependency subtree (a virtual Cargo workspace root emits member entries only). Detectors without per-module emission (sbt, mix, yarn classic, pub, and the node *native* detectors) keep one merged root manifest.
```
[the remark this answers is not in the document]
bomly-guy: Fixed in 1a5cc59 — docs/SCAN_TARGETS.md now lists the gradle module-discovery limitations: literal include(...) declarations (comment-aware) and literal projectDir overrides only, composite builds (includeBuild) not expanded, and the scope-preserving root-only fallback when the multi-project report fails.

internal/detectors/gradle/modules.go
```diff
@@ -0,0 +1,231 @@
...
+				if body[i] == '*' && i+1 < len(body) && body[i+1] == '/' {
+					i += 2
+					break
+				}
+				if body[i] == '\n' {
+					out.WriteByte('\n')
+				}
+				i++
+			}
+		case c == '\'' || c == '"':
```
bomly-guy, line 83: **[P1] Track Groovy/Kotlin triple-quoted strings.** This scanner preserves multiline string contents, so line anchoring does not prevent a line inside `"""..."""` or `'''...'''` from being parsed as a declaration. I reproduced `val sample = """\ninclude(":ghost")\n"""\ninclude(":real")`: `gradleIncludedProjectPaths(stripGradleComments(body))` returns both `:ghost` and `:real`. The phantom task can fail the multi-project invocation and force root-only output. Please track triple-quoted string state (and add a fixture) so declarations inside multiline strings are ignored.

sdk/coordinates.go
```diff
@@ -16,6 +16,14 @@ type Coordinates struct {
...
 	Version        string         `json:"version,omitempty"`
 	Language       Language       `json:"language,omitempty"`
+
+	// FirstParty marks a node as the scanned project's own artifact (its root
+	// package, a workspace member, a reactor module) rather than a consumed
+	// third-party package. Only detectors set it, when synthesizing or matching
+	// nodes they know belong to the build itself; imported documents (SBOM
+	// scans) never do, because a component's declared type is an artifact kind,
+	// not proof of ownership. Enrichment skips first-party nodes.
+	FirstParty bool `json:"first_party,omitempty"`
```
bomly-guy, line 26: **[P1] Mark non-application detector roots too.** The rollout currently updates application-typed roots, but the Go detector's main-module node (`internal/detectors/gomod/detector.go:218-221`) is untyped and never gets `FirstParty`. I reproduced a `github.com/acme/app` main module remaining `NodeIsEnrichable == true` with canonical PURL `pkg:golang/github.com/acme/app`; Scorecard can therefore query the scanned repository, contrary to this feature's root-package contract. Please mark the Go main module (and audit other synthetic non-application roots such as pip-inspect/Pipenv) when they are created.

bomly-guy (commented):
Two remaining correctness gaps reproduced against the latest head.

internal/detectors/gradle/modules.go
bomly-guy: Fixed in b7d51f0. `stripGradleComments` now tracks triple-quoted string state: Kotlin `"""raw"""` and Groovy `'''...'''` strings are blanked down to their newlines (line structure preserved for the anchored regexes; no legitimate declaration lives inside one), so your repro yields only `:real`. Covered by the exact reproduction plus Groovy, unterminated, and `"" + "b"` empty-string-pair edges in `TestStripGradleComments`, and a `"""…include(":ghost")…"""` block in the settings-walk fixture.

sdk/coordinates.go
bomly-guy: Fixed in 4cef443. Audited every detector root-creation site and marked the untyped synthetic roots at creation: the Go main module (your repro — Scorecard can no longer query the scanned repo), the pip-inspect synthetic root (literally named `root`, a real PyPI package), the Pipenv project-typed root, the uv.lock editable project package, every maven TGF block root (single-module artifact + aggregator pom — sibling-consumed reactor modules were already marked via coordinate matching), and gradle's unknown-project placeholder tokens (composite-build members). Assertions added in the gomod, pip-inspect, maven TGF, and uv fixture tests, each also pinning that fetched dependencies stay enrichable. Enrich smoke goldens verified drift-free (the roots had no matched data to lose).

internal/detectors/gradle/main_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	srcPath := filepath.Join(srcDir, "main.go")
+	if err := os.WriteFile(srcPath, []byte(source), 0o644); err != nil {
+		return err
+	}
+	buildCmd := system.Command("go", "build", "-o", outputPath, srcPath)
+	buildCmd.Env = append(os.Environ(), "GOFLAGS=-modcacherw")
+	if buildOutput, err := buildCmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("go build failed: %w (%s)", err, string(buildOutput))
+	}
```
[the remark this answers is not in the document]
bomly-guy: Fixed in b5fa90f. testutil now exposes `BuildGoBinaryFromSource` (same build path as `BuildGoBinary`, self-managed temp source dir) for TestMain callers that have no `testing.TB`, and the gradle fakes build through it.

PR #286, opened 2026-07-17
ci(smoke): parallel per-ecosystem goldens regeneration + build-tool subprocess timeouts

bomly-guy:
Follow-up to #284 (the workflow-parallelization commit there was pushed after the PR merged, so it lands here instead, together with the product-side timeout follow-up).

## Commit 1 — parallel Update Smoke Goldens workflow

The goldens workflow ran the whole smoke suite in one job; a single stalled scan killed everything. `mvn dependency:tree` against a cold `~/.m2` stalled three separate times on 2026-07-17 (runs [29571173463](https://github.com/bomly-dev/bomly-cli/actions/runs/29571173463), [29573048390](https://github.com/bomly-dev/bomly-cli/actions/runs/29573048390), and the test run below). Actions minutes are no longer a concern now that the repo is public, so:

- Regeneration is split into **21 matrix slices** — one per ecosystem, with the slow network-heavy reachability cases isolated — each uploading its changed goldens as an artifact.
- A final `open-pr` job (`if: always()`) merges all artifacts and opens the goldens PR **even when slices fail**, listing failed slices in a PR-body warning. Failed slices are individually re-runnable.
- Slice patterns verified by simulation against a stub test tree using go's real `-run` matcher: every golden is produced by exactly one slice (no gaps, no overlaps).
- Both workflows invoke `go test` directly instead of `make smoke ARGS=...`: slice patterns use `$` anchors that make would expand (`$|` → empty).
- Nightly Smoke coverage gaps closed: `scan-bun`, `scan-recursive-monorepo`, `scan-java-maven-reachability`, `TestPluginWorkflows`, and `TestLiteVersion` were matched by no slice and silently never ran.

**End-to-end test drive**: [run 29574448939](https://github.com/bomly-dev/bomly-cli/actions/runs/29574448939) (workflow taken from this branch, sources from main). 20/21 slices green; `scan-maven` stalled again in the java slice and was cut off cleanly at 5m by #284's harness cap; the other four java goldens still uploaded, and the automation PR (#285, since merged) opened with the failed-slice warning rendered.

## Commit 2 — build-tool subprocess timeouts (product)

The same hazard exists for real users: the pipeline context has no deadline, so a stalled `mvn`/`gradle`/`sbt` hung `bomly scan` forever. All five build-tool invocations (maven resolve + install, gradle resolve + install, sbt dependencyTree) now derive a bounded context via `detectors.BuildToolContext` (10 minutes — generous; healthy cold-cache resolutions finish in 1–2 minutes), mirroring the existing `javaReadyTimeout` pattern. `DeadlineExceeded` is surfaced in the wrapped error and failures flow through the existing degrade-to-fallback paths.

Also documents the slice-matrix contract in `CLAUDE.md`/`AGENTS.md` so new smoke tests get registered in both workflow matrices with exactly-one-slice golden ownership.

## Commit 3 — Maven/Gradle caching

The three mvn stalls all happened during cold-cache downloads from Maven Central. Java derives `user.home` from passwd rather than $HOME (verified: `mvn help:evaluate -Dexpression=settings.localRepository` under an overridden HOME still resolves the real `~/.m2`), so every bomly invocation already shares the runner-level `~/.m2` despite the harness per-test HOME override. Both workflows now cache `~/.m2/repository` (maven slices) and `~/.gradle/{caches,wrapper}` (java slice) with rolling keys — unique per run with prefix restore, since smoke scans clone pinned repos at runtime and there are no lockfiles in the workspace to hash. Warm caches shrink the stall window to near zero and speed the java slices up substantially.

## Commit 4 — Maven mirror: the actual root cause fix

The four `scan-maven`/`scan-java-maven-reachability` stalls on 2026-07-17 trace to transitive POMs of the pinned `example-java-maven` demo (Spring 3.1.1, Keycloak 1.8.1, old slf4j/plexus/neo4j parents) declaring **long-dead repositories** — `snapshots.repository.codehaus.org`, `repository.jboss.org/maven2`, `static.springframework.org`, `oss.sonatype.org` snapshots, and friends. Maven consults them during cold-cache resolution; on the runner network path one accepts connections and never responds, and Maven's default read timeout is effectively unbounded. The maven slices now write a `settings.xml` with `mirrorOf: *` pointing at Central, so dead-repo lookups become fast 404s instead of hangs. Verified: a cold resolve under the mirror succeeds with a **byte-identical dependency tree**.

## Commit 6 — degradation guard (from the #287 incident)

Goldens PR #287 exposed a silent-degradation path: the 5s java readiness probe flaked on a loaded runner, the scan silently fell back `maven-detector` → `syft-detector`, and `-update` wrote the degraded output as a golden. Fixes: `javaReadyTimeout` widened 5s → 30s (only reached when java exists on PATH, so free in the no-java case — this was also the known local flake in the maven/gradle/sbt Ready tests), and the smoke harness now **refuses to compare or write** any output containing a "timed out" reason. Deterministic fallbacks ("executable not found", "exit status 1") remain legitimate golden content; timeout-induced degradation now fails the slice loudly. #287 is closed — main's healthy golden was still current.

## Commit 7 — detector pinning (`--detectors`)

Every scan smoke case now pins the detector its golden encodes: the 22 shared targets via a new `detectors` field on `benchmark.Target` (appended to `SmokeArgs` as `--detectors`), and the 13 explicit cases inline. The engine skips fallbacks outside the pinned selector set (`fallbackSelected` in `pipeline_resolve.go`), so degradation becomes a hard failure: with `--detectors maven` and a broken java runtime the scan exits 3 with `detector maven-detector: not ready` — verified against the unpinned scan which silently emits `syft-detector` output. Cases whose golden legitimately encodes a fallback chain pin both links (`sbt-native,sbt`, `swiftpm-native,swiftpm`).

This audit also caught a **second baked-in degradation**: #285's `scan-maven-multimodule` golden was syft-shaped (`mvn` exit 1 on the dead-repo network path, pre-mirror) — its pre-drift shape was pure maven-native. Regenerated healthy in this PR; pins verified output-neutral against 10 other existing goldens.

## Commit 8 — pin Maven 3.9.16 in CI

The pin-validation run promptly proved the pins' worth: the java slice failed **loudly** with `detector maven-detector: run maven dependency tree: exit status 1` instead of writing a syft golden. Root cause reproduced in an `ubuntu:24.04` container: apt's Maven is **3.8.7**, whose `dependency:tree` cannot resolve in-reactor sibling modules (`Could not find artifact dev.bomly.example:core:jar:1.0.0`) — this is what had been silently degrading the multimodule scan all along. Maven **3.9.16** resolves reactor siblings cold (verified in-container with the mirror, exit 0), so both workflows now install a checksum-verified 3.9.16 tarball instead of apt maven.

## Commit 9 — build once, share across the matrix

Every matrix job compiled the full CLI in `TestMain` (plus the lite binary on demand) — 21 redundant compiles per run. Both workflows now run `make build` once in a `build` job and share `bin/` as an artifact; slices point the harness at the prebuilt binaries via `BOMLY_SMOKE_BINARY` / `BOMLY_SMOKE_LITE_BINARY` (harness falls back to building when unset, so local `make smoke` is unchanged; verified end-to-end locally through the env-var path).

## Verification

- `actionlint` clean on both workflows; slice coverage simulation: none missing, none duplicated.
- `go build ./...`, `go vet`, new unit tests for `BuildToolContext` pass; `make test` green apart from the known java-ready parallel-run flake (passes in isolation, pre-existing).
- Live workflow test drive above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Added a 10-minute safeguard for Maven, Gradle, and SBT operations, preventing stalled builds from running indefinitely.
  - Improved timeout error messages so interrupted build-tool operations clearly indicate the cause.

- **Tests**
  - Expanded smoke-test coverage across package managers, Java build scenarios, plugins, and version checks.
  - Improved automated golden-file regeneration and reporting for smoke-test failures.


bomly-guy:
Adopted all three CodeRabbit findings in e66f404, and went one step further on isolation:

- **smoke.yml now mirrors the goldens matrix exactly**: 21 slices, dedicated reachability slices, `$`-anchored patterns.
- Added a `TestScan$` **top-level anchor** in both matrices — without it, unanchored `TestScan` made `TestScanRecursive*` run in every slice that mentioned it; they now run exactly once (plugin slice).
- Goldens go slice names its `TestAuditDiffAndExplain`/`TestLiteDiff`/`TestLiteExplain` subtests explicitly.
- Golden staging uses `--diff-filter=ACMR`.

Re-verified by simulation with go's real `-run` matcher: every golden owned by exactly one goldens slice, every smoke test in exactly one smoke slice, zero overlaps in either matrix (`TestExamplePluginFixtureCompiles` is intentionally absent — it is untagged and runs in `make test`).

PR #288, opened 2026-07-18
Centralize vulnerability remediation enrichment

bomly-guy:
## Summary

- Add one central, read-only remediation component for enriched vulnerabilities.
- Derive package fix status, a safe recommended version, and occurrence-specific actions after vulnerability consolidation.
- Let built-in and external detectors optionally contribute validated package-manager strategies and advice.
- Project the canonical context through JSON, compact MCP responses, TUI details, text, and Markdown.

## Why

Remediation decisions previously existed only inside the compact MCP projection, while package enrichment exposed a separate version summary. This change gives every surface the same source of truth. MCP and detectors no longer own final action selection.

## Behavior

The enrichment tail is:

```text
consolidate vulnerabilities
  → derive package status and version
  → collect detector advice
  → derive occurrence suggestions
```

Canonical actions are:

- `direct-bump`: update a package declared directly in the project.
- `transitive-override`: pin an indirect package with the package manager override feature.
- `lockfile-refresh`: ask the package manager to resolve a newer indirect package version.
- `no-fix-upstream`: report that the advisory source has no fixed version.
- `manual-review`: report that available data does not support a specific change.

Unknown-parent and non-registry occurrences remain manual. Detector failures, invalid hints, and missing capabilities also fall back to manual review. A recommended version must be stable, comparable, and newer than the installed version.

Remediation applies only to vulnerabilities. License and package-policy findings remain audit results and do not create remediation suggestions.

## Detector and plugin contract

- Detector descriptors may advertise supported package managers and remediation strategies.
- Optional providers receive cloned detection and package-registry data.
- Hints may identify supported strategies and plain-language manager advice.
- Core strips control characters, limits advice length, and caps detector warnings across subprojects.
- Hints cannot choose a version or final action, edit files, run commands, or make network calls.
- Each detector owns its capability declaration and advice implementation.
- The protocol-v1 RPC is additive. Plugins that do not advertise the capability are not called and continue to work unchanged.

## Output

- Scan and diff JSON expose remediation under `packages[]`.
- Explain JSON filters suggestions to the focused dependency and does not repeat them in dependency paths.
- `affected_dependency_refs` identifies occurrences of the vulnerable package. `suggested_action_dependency_ref` identifies the dependency or manifest anchor the suggestion applies to.
- MCP projects canonical suggestions while preserving its ranking, size caps, policy information, and `mcp/1` schema.
- Vulnerabilities allowed by audit policy or below the failure threshold remain visible as suppressed information, not actionable groups.
- TUI component details show a dedicated Remediation Suggestion section directly below Vulnerabilities, with one blank row between sections. Human-facing status labels say Complete fix available, Partial fix available, No fix available, or Fix availability unknown. These labels describe advisory evidence, not a verified project change.
- Text output shows only a short fix summary directly below enrichment and points to JSON for details.
- Markdown shows the same summary and the complete suggestion table, with readable package-manager advice.
- Text and Markdown summaries count only concrete bump, override, and refresh actions backed by complete fix evidence. Manual review, no-fix, partial, and unknown guidance remains in detailed output without inflating the headline.
- SARIF remains unchanged.

CLI JSON remains schema `1.0`; MCP remains `mcp/1`; plugin protocol remains `v1`.

## Safety

- No `--fix` flag or write path.
- No manifest or lockfile mutation.
- No package-manager execution for remediation.
- No additional network calls beyond requested enrichment.
- Synthetic manifest roots never count as safe parent evidence.

## Validation

- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- Focused SDK, derivation, MCP, TUI, scan, diff, explain, Markdown escaping, empty-state, summary-count, and section-spacing tests.
- The pinned Go reachability smoke case passes.
- The npm live slice reaches golden comparison, where current advisory data changes the unrelated unreachable count from 24 to 25; that drift is not included here.
- Maven cannot run locally because Java is unavailable. Python cannot run locally because `pip` is unavailable. No runtime or package manager was installed.
- `git diff --check`


bomly-guy:
@coderabbitai review

bomly-guy:
@claude review

sdk/vulnerability.go
```diff
@@ -548,13 +548,72 @@ type Finding struct {
 	// VulnerabilityID is the advisory id within the referenced package, set
 	// for vulnerability-kind findings.
 	VulnerabilityID string `json:"vulnerability_id,omitempty"`
+	// OverrideEvidence carries occurrence-specific, read-only package-manager
```
bomly-guy: **SDK model — override evidence.** Findings can now carry the exact package-manager declaration that produced an override-hygiene result. `OverrideEvidence` keeps manager/version, manifest and workspace ownership, selector/container details, scopes, source, locations, and materialized graph occurrences together. That occurrence-specific shape is important for duplicate versions and workspaces: neither the auditor nor remediation has to infer ownership from a flattened package record.

sdk/detector.go
```diff
@@ -100,6 +100,10 @@ type DetectorDescriptor struct {
 	PackageManagerSupport []PackageManagerSupport `json:"packageManagerSupport,omitempty"`
 	FallbackDetectors     []string                `json:"fallbackDetectors,omitempty"`
 	SupportsInstallFirst  bool                    `json:"supportsInstallFirst,omitempty"`
+	// RemediationCapabilities advertises optional, read-only fix hint support.
```
bomly-guy: **SDK extension point — capability advertisement.** Remediation support is additive to `DetectorDescriptor`; an empty/omitted list means the detector behaves exactly as before. This avoids introducing another plugin kind. A detector opts in only to read-only hint production, while core retains candidate selection, safety policy, validation, and write authority.

sdk/detector.go
```diff
@@ -114,6 +118,65 @@ type DetectorDescriptor struct {
...
+
+// RemediationHintRequest supplies completed detection and enrichment evidence
+// to an optional detector hint provider.
+type RemediationHintRequest struct {
+	ProjectPath string           `json:"projectPath,omitempty"`
+	Detection   DetectionResult  `json:"detection"`
+	Registry    *PackageRegistry `json:"registry,omitempty"`
+}
+
+// RemediationOperationHint describes a structured, non-executable operation
```
bomly-guy: **SDK model — hints are data, not commands.** A detector can describe a structured operation and its provenance, but this object does not authorize execution. Core validates paths, action type, occurrence ownership, manager ownership, and applicability before translating any hint into an automatic candidate. `Executable` and `Arguments` remain representational fields; the initial automatic path rejects detector-supplied executable operations.

sdk/serve.go
```diff
@@ -76,6 +83,13 @@ type Client interface {
 	Audit(context.Context, *AuditRequest) (*AuditResponse, error)
 }
 
+// DetectorRemediationClient is the additive client capability for optional
```
bomly-guy: **Protocol-v1 compatibility.** The remediation client capability is a separate interface rather than a new method on `Client`. That preserves source compatibility for existing client implementations and test doubles. Core type-asserts this interface only for detectors that advertise remediation support; older servers return `Unimplemented` for the optional RPC and continue to load normally.

internal/remediation/model.go
```diff
@@ -0,0 +1,186 @@
...
+	Evidence        int `json:"evidence"`
+	Candidates      int `json:"candidates"`
+	Groups          int `json:"groups"`
+	Automatic       int `json:"automatic"`
+	Manual          int `json:"manual"`
+	Unavailable     int `json:"unavailable"`
+	Vulnerabilities int `json:"vulnerabilities"`
+}
+
+// Context is the canonical remediation projection derived after enrichment.
```
bomly-guy: **Canonical model boundary.** The model deliberately separates facts from decisions: `FixEvidence` records advisory/occurrence facts, `Candidate` proposes one occurrence-specific action, `Group` coalesces changes that solve multiple findings, and `Diagnostic` explains why automation is unavailable. `Context` is the shared projection consumed by CLI JSON, terminal/Markdown, TUI, SARIF, and MCP. `FixResult` separately records the terminal transaction outcome.

internal/engine/pipeline.go
```diff
@@ -41,9 +43,84 @@ func (p *Pipeline) Run(ctx context.Context, req PipelineRequest) (PipelineResult
 		return result, err
 	}
 	p.runAudit(ctx, &result, req)
+	p.DeriveRemediation(ctx, &result, req)
```
bomly-guy: **Pipeline placement.** Remediation derivation runs after match/enrichment, optional reachability analysis, and optional auditing. `DeriveRemediation` is gated by `EnrichEnabled`, so every enriched scan gets a plan automatically; analysis and audit can add priority or policy evidence but are not prerequisites. Explain and diff call the same derivation entry point after constructing their specialized projections.

internal/remediation/derive.go
```diff
@@ -0,0 +1,855 @@
...
+	candidate := Candidate{
+		Target:             target,
+		AffectedPackageRef: dependency.PackageRef,
+		RecommendedVersion: recommended,
+		EvidenceRefs:       []string{evidenceID},
+		Confidence:         ConfidenceHigh,
+		Availability:       AvailabilityManual,
+	}
+	switch {
+	case relationship == sdk.DependencyRelationshipUnknown:
```
bomly-guy: **Unknown-parent safety boundary.** An occurrence whose dependency relationship cannot be proven is always converted to low-confidence manual review with an `unresolved-parent` diagnostic. The engine never guesses a parent upgrade from a synthetic or incomplete path. Transitive occurrences are also manual initially because an advisory’s fixed child version does not prove which parent release is safe.

internal/remediation/transaction.go
```diff
@@ -0,0 +1,783 @@
...
+		result.Reason = err.Error()
+		return result, fmt.Errorf("apply remediation proposal: %w", err)
+	}
+	result.Phase = "resolution"
+	if err := regeneratePackageManagerState(ctx, disposableRoot, candidates, logger); err != nil {
+		result.Reason = err.Error()
+		return result, fmt.Errorf("regenerate package-manager state: %w", err)
+	}
+
+	result.Phase = "validation"
```
bomly-guy: **Transactional fix loop.** `--fix` first applies the complete safe candidate set in a disposable copy, regenerates manager state with lifecycle scripts disabled, and rescans. Only an effective, non-regressive, contained file delta reaches the real checkout. The real checkout is rescanned again and must match the validated result; any failure or divergence restores backups and reports the phase and reason.

bomly-guy (commented):
Guided tour of the remediation architecture, anchored to the current PR head. A useful reading order is: shared SDK evidence → optional detector contract → protocol compatibility → canonical remediation model → pipeline placement → safety classification → transactional application.

bomly-guy:
Status update: this draft is intentionally deferred.

We want to stabilize existing capabilities and land smaller, bounded improvements before introducing the remediation pipeline. Please do not merge or spend review time on this PR for now. The branch will remain open as implementation evidence and a design reference, but it will not be rebased or expanded while deferred.

When the work resumes, we will reassess the contracts against then-current `main`, rebase or reconstruct the implementation as appropriate, and rerun the full generation, test, build, lint, compatibility, output, and transactional safety suites before making a merge decision.

internal/output/sarif.go
```diff
@@ -224,25 +223,20 @@ func WriteSARIF(w io.Writer, findings []sdk.Finding, registry *sdk.PackageRegist
...
-			}}
-		}
 		props := sarifPropertiesFromFinding(f, locationURIs)
 		if vuln != nil {
 			props = mergeSARIFProperties(props, sarifPropertiesFromVulnerability(vuln, includeReachability))
 			if includeReachability && vuln.Reachability != nil && len(vuln.Reachability.CallPaths) > 0 {
 				result.CodeFlows = buildSARIFCodeFlows(vuln.Reachability.CallPaths)
 			}
 		}
+		props = mergeSARIFProperties(props, sarifRemediationProperties(f, remediationFromSARIFOptions(options)))
```
bomly-guy: **[critical] This new merge call clobbers `kev_exploited` and `reachability_dynamic_imports_detected` to `false` on every SARIF result.**

`mergeSARIFProperties` unconditionally assigns `base.KEVExploited = extra.KEVExploited` and `base.DynamicImportsDetected = extra.DynamicImportsDetected`. Before this PR the function ran once (finding + vulnerability props) and the overwrite was harmless; this second call's `extra` (remediation-only props) always carries both booleans as `false`, so a vulnerability that was marked KEV-exploited loses `"kev_exploited": true` in every SARIF export — remediation context present or not. KEV is a primary triage signal for code-scanning consumers, and the new `remediation_test.go` never sets those fields so nothing catches it.

Guard those two assignments (only overwrite when `extra` is true / when merging vulnerability props), and add a SARIF test that sets `KEVExploited: true` alongside a remediation context.

internal/cli/fix_transaction.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		defer prepared.Close()
+		pipeline := engine.NewPipeline(prepared.Registry(), logger)
+		request := prepared.PipelineRequest(scope, stderr)
+		result, err := scanengine.Run(ctx, pipeline, request)
+		if err != nil {
+			return remediation.ValidationSnapshot{}, err
+		}
+		pathAbsolute, _ := filepath.Abs(path)
+		if pathAbsolute == projectAbsolute {
```
bomly-guy: **[major] On any symlinked project path (every macOS `/tmp`, `/var`), the post-fix rescan is silently dropped and the user sees the stale pre-fix report.**

`ApplyTransaction` resolves the root through `filepath.EvalSymlinks` (transaction.go:67-69) and passes the *resolved* path to the final `Rescan` (transaction.go:174), but this closure compares `filepath.Abs(path)` against the *unresolved* `projectAbsolute` from line 25. When they differ, `finalResult` stays nil and `scan_cmd.go` falls back to the initial pipeline result — the just-fixed vulnerabilities are still reported as open even though files were changed. `transaction_test.go:74` works around this with `EvalSymlinks`, but the CLI glue doesn't.

Apply `EvalSymlinks` to `projectAbsolute` too (or compare resolved-to-resolved), and stop discarding both `filepath.Abs` errors with `_` — either failure silently produces the same stale-report outcome.

bomly-guy (commented):
Full review of the remediation pipeline at head `e07600c`. Overall the architecture is sound and several safety claims verified cleanly: subprocess argv is fully constant (package names/versions never reach command arguments, so no `-`-prefix injection), executables come from a hard whitelist via `LookPath`, detector hints are unconditionally demoted to manual and re-validated before any write, protocol-v1 compatibility is genuinely additive (old plugins pass validation; `sdk/remediation_plugin_test.go` proves the `Unimplemented` path), MCP is force-read-only with a locking test, `make generate` output is drift-free, go.mod is untouched, and the decision log entry is present. The `restore explain finding reachability` follow-up commit is complete — no other dropped assignments in the first commit.

That said, there is one **critical** output regression (SARIF `kev_exploited` clobbering), several **major** correctness/safety issues flagged inline, and cross-cutting gaps below.

### Checklist gaps (blocking per repo checklist)

- **Smoke coverage is entirely missing.** No `test/smoke/*` or smoke-workflow changes. Because `DeriveRemediation` sets a non-nil context on every enriched run and `view.go` emits it via `omitempty` on a pointer, **every existing enrich golden breaks post-merge** — zero goldens currently contain a `remediation` key, and `normalizeJSON` has no scrubbing for the new volatile fields (`recommended_version` drifts as advisories publish, so even regenerated goldens will flake without a normalizer). Also no smoke case exercises `--fix` end-to-end (checklist requires a pinned real repo), and no slice registration in `smoke.yml` / `update-smoke-goldens.yml`.
- **CLAUDE.md drift**: the "Scan pipeline order" line was not updated with the remediation stage (AGENTS.md and both ARCHITECTURE docs were), and the Package Boundaries list doesn't cover the new `internal/remediation` / `internal/overrides` packages (engine and registry now import `remediation`; both new packages depend only on `sdk` — worth pinning that down as a rule).
- **`docs/auditors/override-hygiene.md`** (generated) misdescribes the auditor: "Finding kind: policy / policy match", but the auditor emits `Kind: sdk.FindingKind(item.ReasonCode)` with reasons like `override-ineffective-nesting` / `phantom-import`. Fix belongs in the descriptor/prose metadata the generator reads.
- **`docs/MCP.md:121-124`**: the new read-only paragraph is inserted mid-bullet-list (splits the "For the omitted detail:" list when rendered).

### Issues on lines outside the diff (can't anchor inline)

- **`internal/cli/mcp_cmd.go:375-379`** (minor, leak): `RunScan` gained `defer cmdCtx.Close()`, but `RunExplain` still calls `o.Prepare(...)` without ever closing the prepared context — every MCP explain call leaks temp clone dirs for remote targets. `RunDiff` closes its targets, so explain is now the only leaking path.
- **`internal/mcp/compact_diff.go:93-96`** (minor): with a canonical head context, `buildRemediations` ignores the `headInput.Findings = open` scoping — the context is derived from the full head registry irrespective of audit dispositions, so vulnerabilities suppressed via `allow_vulnerability_ids` can reappear as diff remediation groups, contradicting the comment above that line.
- **`internal/remediation/transaction.go:555`** (minor): `higherVersion` returns its left argument when parsing fails, so a graph node with an unparseable resolved version passes the `>= RecommendedVersion` check in `targetExistsInGraph` and the upgrade is "verified" without proof.
- **`internal/remediation/derive.go:747-769`** (minor): `lowestFixedVersionAbove` with an unparseable current version returns the lowest parseable fixed version unconditionally (can recommend below the version in use; manual candidates only), and `compatibleAutomaticUpgrade` accepts a **prerelease** fixed version within the same major — auto-pinning users to an RC is arguably not a safe automatic change.
- **`internal/tui/posture_diff.go:821`** (nit): stale "1-7 switch tabs" help text — the diff shell now has 8 tabs; every other diff tab's `filterHelp` was updated.
- **`internal/overrides/source.go:84-88`** (minor): phantom-import evidence hardcodes `Manager: sdk.PackageManagerNPM` regardless of the project's real manager and stores a source-file path in `ManifestPath` — surprising field semantics for downstream consumers of the sdk shape.

### Test-coverage gaps worth closing

- Only a fake `npm` is ever executed; pnpm/yarn/bun command construction is untested (would have caught the Yarn Berry flag issue). No tests for ctx cancellation mid-transaction, mid-write rollback, `copyProject` symlink rejection, or `changedFixFiles` unexpected-file rejection. `TestApplyTransactionDisablesLifecycleScripts` asserts env-var presence, not script suppression (and `PNPM_CONFIG_IGNORE_SCRIPTS` isn't a variable pnpm reads — pnpm honors `npm_config_ignore_scripts`; harmless since the flag is also passed, but the test cements false confidence).
- No test renders a **populated** Remediation tab in the TUI (scan or diff); only nil-context tab switching is exercised.
- `internal/overrides/reader_test.go` misses exactly the real-world shapes that turn out to be buggy: yarn glob/path selectors, npm `"."` self-reference, version-qualified keys, multi-level `>` chains, `$package` values, workspace-child manifests, malformed/oversized file degradation.

### Smaller notes

- `requireCleanGitWorktree` drops the underlying error (no `%w`) on `git rev-parse` failure, and that invocation lacks the DEBUG command-line log the `status` call has.
- Failed package-manager `CombinedOutput` is discarded (only `output_bytes` logged) — log it at DEBUG so `-vv` users can diagnose lockfile regeneration failures.
- Base-side remediation derivation in `internal/engine/diff/diff.go:87` is dead work (nothing consumes `Base.Remediation`), and on `--enrich --audit` runs `overrides.InspectWithOptions` walks the tree twice (auditor + derive).
- Fix-failure payload is written only for primary-format JSON; `-o json=file`, markdown, and SARIF specs get nothing, and `exit.ResolutionFailureError` conflates fix failure with resolution failure.
- SARIF remediation properties match only via `Evidence` (PackageRef+VulnID), so policy findings linked by `FindingRefs` never get remediation properties, and there's no `f.ID` fallback like MCP's `findingVulnID`.
- `sarifRemediationProperties`' variadic `remediationContexts ...*remediation.Context` on `render.Scan` (`internal/cli/render/scan.go:39`) has one caller and reads only index 0 — a plain pointer parameter would be clearer.
- CLAUDE.md's network allowlist ("Network calls (--enrich) permitted only to: ...") doesn't cover the package registries that lockfile regeneration necessarily contacts during `--fix` — the contract needs amending to stay honest.

internal/remediation/transaction.go
```diff
@@ -0,0 +1,783 @@
...
+}
+
+func packageManagerCommand(manager string) (string, []string, error) {
+	switch manager {
+	case "npm":
+		return "npm", []string{"install", "--package-lock-only", "--ignore-scripts", "--no-audit", "--no-fund"}, nil
+	case "pnpm":
+		return "pnpm", []string{"install", "--lockfile-only", "--ignore-scripts"}, nil
+	case "yarn":
+		return "yarn", []string{"install", "--ignore-scripts"}, nil
```
bomly-guy: **[major] `yarn install --ignore-scripts` is not a valid Yarn Berry (v2+) option, so `--fix` always fails at the resolution phase for Berry projects — the advertised Yarn support is effectively yarn-classic-only.**

Berry's CLI rejects unknown options, so the transaction dies before `YARN_ENABLE_SCRIPTS=false` ever matters (fail-closed, but the feature is dead for most modern yarn projects). Berry needs `--mode=update-lockfile` plus the env var instead. On classic yarn there is no lockfile-only mode, so the disposable copy performs a **full network install** materializing the entire tree — heavier and a much wider network surface than "regenerates package-manager state" suggests.

Either branch on the detected yarn major, or scope the claim in `docs/REMEDIATION.md:29` to yarn classic and document the full-install behavior. Note no test executes the pnpm/yarn/bun command lines (only a fake `npm`), which is why this wasn't caught — worth at least a unit test asserting the constructed argv per manager against known-valid flags.

internal/remediation/derive.go
```diff
@@ -0,0 +1,855 @@
...
+		packageJSON := manifestPath
+		if filepath.Base(packageJSON) != "package.json" {
+			packageJSON = filepath.Join(filepath.Dir(packageJSON), "package.json")
+		}
+		lockfilePath, ok := managerLockfile(projectRoot, manifestPath, packageJSON, candidate.Target.PackageManager)
+		if !ok {
+			candidate.Limitations = appendUnique(candidate.Limitations, "an existing manager-owned lockfile could not be proven for transactional regeneration")
+			continue
+		}
+		section, specifier, ok := declaredDependency(packageJSON, candidate.Target.Name)
```
bomly-guy: **[major] The declared specifier is never compared against the recommended version, so a stale committed lockfile makes the auto-fix rewrite the declaration *downward*.**

`declaredDependency` only checks the name is declared and the specifier isn't an unsafe protocol; the compatibility gate on line 129 compares `Target.CurrentVersion` (the *locked* version) to `RecommendedVersion`. If `package.json` declares `"example": "^2.0.0"` but the committed lockfile still pins 1.0.0 and the advisory fix is 1.2.0, the transaction rewrites the declaration to `"1.2.0"` — a silent semantic downgrade of the user's declared range. A clean git worktree doesn't prevent this; out-of-sync committed lockfiles are common.

Related (minor): the applied edit always replaces a range (`^1.0.0`) with an exact pin (`1.2.0`), losing the caret/tilde style with no mention in the report. Consider parsing the declared range and skipping (or demoting to manual) any candidate whose recommended version doesn't satisfy it, and preserving the range operator on rewrite.

internal/remediation/transaction.go
```diff
@@ -0,0 +1,783 @@
...
+		}
+	}
+	for path, edits := range editsByPath {
+		fullPath := filepath.Join(root, filepath.FromSlash(path))
+		data, err := os.ReadFile(fullPath)
+		if err != nil {
+			return fmt.Errorf("read %s: %w", path, err)
+		}
+		var document map[string]any
+		if err := json.Unmarshal(data, &document); err != nil {
```
bomly-guy: **[major] Round-tripping `package.json` through `map[string]any` + `MarshalIndent` rewrites the entire file: keys are alphabetized, original ordering/formatting is destroyed, duplicate keys silently collapse, and numbers re-encode via float64.**

For a one-line version bump the user gets a full-file restructuring diff (`name`/`version` no longer first, `scripts` reordered, custom indentation lost). That contradicts the surgical-delta framing and churns every subsequent human review of the manifest. A targeted byte-level edit of the single declaration value (or an order-preserving JSON editor) is needed before this is safe to run on real projects.

internal/cli/opts/options.go
```diff
@@ -188,6 +183,9 @@ func (o *Options) ResolveConfig(cmd *cobra.Command) error {
 	if flagChanged(cmd, "max-depth") && !resolved.Recursive {
 		return exit.InvalidInputError("--max-depth requires --recursive")
 	}
+	if resolved.Fix && cmd.Name() != "scan" {
```
bomly-guy: **[major/security] A repo-committed `.bomly/config.yaml` (or `BOMLY_FIX=1`) can turn a plain `bomly scan` into filesystem mutation plus package-manager subprocess execution — no CLI flag or per-run consent required.**

Project config auto-loads from the scanned directory, and `pipeline: {enrich: true, fix: true}` passes all validation. So `git clone untrusted-repo && bomly scan` can rewrite manifests and shell out to package managers (a fresh clone trivially satisfies the clean-worktree gate; on classic yarn that's a full network install). For a security-sensitive scanner routinely pointed at untrusted checkouts, an explicit-consent action shouldn't be reachable from data inside the tree being scanned.

Suggest honoring `Fix` only when set on the command line — the `flagChanged(cmd, "max-depth")` gate two lines above is exactly the right precedent. Related (minor): as written, `fix: true` from config/env also breaks every non-scan command (`explain`, `diff`, `mcp`, `version` all die here with a message that says "--fix" even when no flag was passed), with no per-invocation escape hatch since non-scan commands don't bind the flag.

internal/registry/builder.go
```diff
@@ -422,6 +435,7 @@ func (r *Registry) registerAuditors() {
 			TyposquatThreshold: threshold,
 			TyposquatMode:      r.configs.TyposquatMode,
 		},
+		overridehygiene.Auditor{},
```
bomly-guy: **[major] Registering the override-hygiene auditor unconditionally makes `bomly diff --audit` emit false `override-orphaned-target` findings for npm projects with overrides.**

Diff audits run against *focused* graphs containing only added/removed/changed dependencies (`focusedAuditGraphs` in `internal/engine/diff/diff.go`), but `overridehygiene.Auditor.Audit` inspects the full project tree and emits `override-orphaned-target` whenever the graph has no matching occurrence. Any override whose target dependency is unchanged between refs is absent from both focused graphs, so every such override yields a spurious introduced/persisted finding on both sides of the diff. Scan/explain are unaffected (full graphs).

The auditor needs either the full graph in diff mode or a way to opt out of focused-graph audits.

internal/overrides/reader.go
```diff
@@ -0,0 +1,450 @@
...
+
+func selectorParts(selector, inheritedParent string) (string, string) {
+	selector = strings.TrimSpace(selector)
+	if before, after, ok := strings.Cut(selector, ">"); ok {
+		return selectorName(before), selectorName(after)
+	}
+	return inheritedParent, selectorName(selector)
+}
+
+func selectorName(selector string) string {
```
bomly-guy: **[major] Selector parsing misses the most common real-world key shapes, producing guaranteed false `override-orphaned-target` findings.**

Verified by execution against the current code:
- **Yarn path/glob selectors**: resolutions keys are `/`-separated (`"**/left-pad"`, `"d2/left-pad"`), not `>`. `selectorName` returns the raw string, which never matches a graph node — both documented yarn shapes emit `override-orphaned-target` even when `left-pad` is materialized (and `path.Match("**/left-pad", ...)` can only ever match scoped names since `*` doesn't cross `/`).
- **npm `"."` self-reference**: `{"overrides":{"foo":{".":"1.0.0","bar":"2.0.0"}}}` is documented npm syntax (override `foo` itself) but yields target `"."` and a false orphan.
- **Multi-level chains**: `"a>b>c"` cuts at the first `>`, producing target `"b>c"` which can never match.
- **Version-qualified keys** (minor): `"foo@^1.0.0": "1.0.5"` strips the range and matches *every* materialized `foo`, so out-of-range occurrences trigger false `override-surpassed-pin`.

Given `override-orphaned-target` is a user-facing warning finding, these shapes need parsing (or at minimum exclusion from the orphan check) before the auditor ships enabled-by-default. `reader_test.go` covers none of them.

internal/overrides/reader.go
```diff
@@ -0,0 +1,450 @@
...
+		return nil, nil // malformed project input degrades without aborting audit
+	}
+	manager, managerVersion := detectManager(filepath.Dir(filePath), document)
+	relative, _ := filepath.Rel(root, filePath)
+	workspace := ""
+	if filepath.Dir(relative) != "." {
+		workspace = filepath.ToSlash(filepath.Dir(relative))
+	}
+	var declarations []declaration
+	containers := []string{"overrides"}
```
bomly-guy: **[major] pnpm's canonical `"pnpm": {"overrides": {...}}` block in package.json is never read — only top-level `overrides`/`resolutions` keys are inspected.**

pnpm projects using the standard pre-pnpm-10 layout (still supported) get zero override evidence, while a *top-level* `overrides` key in a pnpm project is flagged `override-wrong-container` — so the checker both misses the correct location and can't tell users where the correct one is. `pnpm-workspace.yaml` is handled, but the package.json `pnpm` block is the more common single-package shape.

internal/overrides/reader.go
```diff
@@ -0,0 +1,450 @@
...
+		vulnerable, stale := vulnerableOverrideState(occurrences, item.requested, registry)
+		if vulnerable {
+			appendReason("override-materialized-vulnerable", "enrichment")
+		}
+		if stale {
+			appendReason("override-stale-floor", "enrichment")
+		}
+	}
+	if options.AnalysisEnabled {
+		phantoms, phantomErr := inspectPhantomImports(root, graph)
```
bomly-guy: **[major] Any single stat/read failure during inspection aborts the whole audit and discards already-computed evidence — inconsistent with the deliberate malformed-JSON tolerance.**

Verified by execution: one broken symlink named `broken.js` makes `InspectWithOptions` return `(nil, error)`, throwing away all offline override findings computed above this line. Same all-or-nothing behavior for one oversized (>2 MiB) `package.json` anywhere in the tree (e.g. a test fixture kills declaration reading for the valid root manifest too), unreadable directories, and repos over the 10,000-file `maxSourceFiles` cap. Since the engine downgrades auditor errors to warnings, the practical effect is the auditor silently yields **zero findings** on large or slightly-messy repos.

Malformed JSON already degrades per-file (`return nil, nil` in `readPackageJSON`); apply the same policy here — skip the offending file/subtree, keep the rest. Also worth noting: every override in a non-root manifest is flagged `override-wrong-container` via `item.workspace != ""`, which false-positives on nested standalone subprojects under `--recursive` (own lockfile, not a workspace member) where the override is fully effective.

internal/mcp/remediation.go
```diff
@@ -96,6 +101,207 @@ func buildRemediations(in remediationInput) remediationOutput {
 	return remediationOutput{Remediations: out, Informational: informational, Truncation: trunc}
 }
 
+func projectCanonicalRemediation(in remediationInput) remediationOutput {
```
bomly-guy: **[major] The canonical projection permanently empties the compact `Informational` channel and drops audit-disposition classification for every real MCP run.**

`projectCanonicalRemediation` never populates `Informational`, and since the pipeline now attaches a canonical context on every enriched run (and `--audit` requires `--enrich`), the legacy path that routed warn-disposition findings to `Informational` is dead code in practice. Consequences in `compact_scan.go`: `Summary.Informational` is always 0, warn-disposition findings are either reclassified as actionable or silently dropped (a vulnerability finding with no matching evidence appears nowhere), and `Summary.Actionable` inflates. The mcp/1 schema fields still exist but can never be filled — either populate them from the canonical context or document the contract change.

Related in this file: `compactAction` folds `ActionParentUpgrade` into `lockfile-refresh` but `finalizeCanonicalGroup` sets `OverrideAdvice` only for `ActionTransitiveOverride`, whereas the legacy path attached advice to lockfile-refresh groups too — agents lose the package-manager-specific override snippet for the standard transitive case (and `_ = canonical` at line 231 flags the dead parameter). Truncation counters are also inexact: groups dropped at the cap don't count `FindingRefs` in `OmittedFindings`, while the per-fix cap counts refs that would fail the `evidenceByID` lookup.

internal/cli/render/remediation.go
```diff
@@ -0,0 +1,99 @@
...
+	if result.Reason != "" {
+		message += " — " + result.Reason
+	}
+	if result.RolledBack {
+		message += " (changes rolled back)"
+	}
+	return message
+}
+
+func fixResultMarkdown(result *remediation.FixResult) []string {
```
bomly-guy: **[major] `fixResultMarkdown` returns `FixResultText`, which embeds raw ANSI escape sequences (`render.Style(..., Green, Bold)` is unconditional) — every `--format markdown` report after a fix contains literal `\x1b[...` bytes.**

Markdown is the format used in PR comments and Guard summaries, and every other markdown helper in this package is ANSI-free; this is the only leak. `fixResultMarkdown` needs a plain-text rendering path.

internal/engine/pipeline.go
```diff
@@ -41,9 +43,84 @@ func (p *Pipeline) Run(ctx context.Context, req PipelineRequest) (PipelineResult
...
+		zap.Int("manifests", len(result.Consolidated.Manifests)))
+	hints := p.remediationHints(ctx, result, req)
+	overrideEvidence, overrideErr := overrides.InspectWithOptions(req.ProjectPath, result.Graph, result.Registry, overrides.InspectOptions{
+		AnalysisEnabled: req.AnalyzeReachabilityEnabled,
+	})
+	if overrideErr != nil {
+		result.RemediationWarnings = append(result.RemediationWarnings, PipelineWarning{Source: "override-hygiene", Message: overrideErr.Error()})
+		p.Logger.Warn("pipeline: override evidence unavailable", zap.Error(overrideErr))
+	}
+	context := remediation.Derive(remediation.Input{
```
bomly-guy: **[major] `result.Remediation` is set unconditionally on every enriched run, which (a) breaks every existing enrich smoke golden and (b) adds a permanent zero-value section to all output surfaces.**

(a) No golden currently contains a `remediation` key, `normalizeJSON` has no scrubbing for the new volatile fields (`recommended_version` drifts as advisories publish), and this PR touches neither `test/smoke/` nor the smoke workflows — nightly smoke goes red on every enrich case post-merge. See review body for the full checklist gap (including the missing `--fix` smoke case).

(b) Every enriched text report now prints `Remediation 0 proposed changes (0 automatic, 0 manual, 0 unavailable)` since `RemediationText` prints its header unconditionally, and markdown reports emit headed-but-empty `## Remediation` / `## Fix Result` sections on unenriched runs (`remediationMarkdown(nil)` returns nil but `writeMarkdownReport`'s `_No details._` placeholder only fires on nil `Lines`). Consider suppressing the section when the context is empty, or emitting the sibling-style placeholder sentence.

Also here: `RemediationWarnings` recorded at lines 64/112 reach only MCP diagnostics and `-v` logs — detector/match/audit warnings all flow to progress ⚠ children (the only default-visible channel), so a failed override inspection is invisible in a default-verbosity run. And nit: `context := remediation.Derive(...)` shadows the imported `context` package inside a function taking `ctx context.Context`.

internal/remediation/transaction.go
```diff
@@ -0,0 +1,783 @@
...
+	return closed, persisted, introduced
+}
+
+func automaticCandidates(context *Context) []Candidate {
+	var candidates []Candidate
+	for _, candidate := range context.Candidates {
+		// The initial mutation surface is intentionally limited to direct
+		// registry declarations. Other actions remain reviewable hints even if
+		// malformed or stale input marks them automatic.
+		if candidate.Action == ActionDirectUpgrade && candidate.Target.Relationship == sdk.DependencyRelationshipDirect && candidate.Target.Source == sdk.DependencySourceRegistry && candidate.AutoApplicable && candidate.Availability == AvailabilityAutomatic {
```
bomly-guy: **[minor] Derivation and transaction disagree on empty `Target.Source`, so the report can advertise automatic candidates the transaction then refuses.**

`annotateLocalManifestOperations` (derive.go:102) accepts `Source == ""` and marks the candidate automatic (counted in `Summary.Automatic`), but `automaticCandidates` here requires `Source == sdk.DependencySourceRegistry` exactly — such a candidate is filtered out and `ApplyTransaction` fails with "no safely actionable remediation changes were available" despite the report claiming automatic fixes exist. Align the two predicates (probably by treating empty as registry in both places, or neither).

While in this area: rollback is only invoked on returned errors — a panic in `request.Rescan` (which runs the full scan pipeline) leaves applied files in place with no rollback and no `RolledBack` record; a `defer`/`recover` around the apply-through-rescan window would close that hole.

bomly-guy:
## Re-review at `655f47f`

I re-verified every finding from my previous review against the new head (each checked in code, most re-verified by executing the new tests or targeted probes). **All 14 inline findings are correctly fixed and their threads are now resolved**, including the critical SARIF `kev_exploited` clobber (now guarded and pinned by a test), the config/env-driven `--fix` consent hole (`Fix` is now flag-only: env tag removed, YAML leaf deleted and rejected, `flagChanged` gate with tests), the Yarn Berry breakage (generation-aware `--mode=update-lockfile`), the `package.json` map round-trip (replaced with a token-offset byte splice that preserves formatting), the stale-lockfile downgrade path (declared-range-aware replacement), the diff focused-graph false positives (override-hygiene excluded from diff audits), the selector-parsing family (`**/pkg`, `"."`, `a>b>c`, `@`-qualified keys — all execution-verified), the pnpm block, per-file degradation, the MCP informational channel, the ANSI-in-markdown leak, and the panic-rollback gap. Most body-level items also landed: the end-to-end `--fix` smoke test with correct `$`-anchored slice registration and golden scrubbing, `RunExplain` context leak, compact-diff finding scoping, SARIF policy-finding properties, generated-docs corrections, CLAUDE.md pipeline/boundary updates, and the network-policy carve-out. `make generate` is drift-free, go.mod untouched, Test/Build/Lint green.

### Still open

1. **CodeQL gate is red** — two `go/allocation-size-overflow` alerts at `internal/remediation/transaction.go:472`: `make([]byte, 0, len(data)-(absoluteEnd-absoluteStart)+len(encoded))` computes capacity from decoder offsets with no bounds guard. Validate `0 <= absoluteStart <= absoluteEnd <= len(data)` before allocating; that clears the gate and hardens the splice against any future offset bug.
2. **Satisfying-range declarations make the automatic fix a guaranteed no-op** — when the declared range already permits the fix (declared `^1.0.0`, locked `1.0.0`, fix `1.2.0` — the most common npm shape), `automaticDeclarationReplacement` returns the original specifier, the manifest edit changes zero bytes, `npm install --package-lock-only` honors the in-sync lockfile, and the transaction fails with "validated remediation produced no file changes". Fail-safe, but the flagship path can't succeed for its most common input (the new smoke fixture sidesteps this with an exact pin). Needs a lockfile-targeted regeneration strategy for this case, or an explicit limitation in `docs/REMEDIATION.md` plus a candidate demotion so the report doesn't advertise an automatic fix that will always fail.
3. **`higherVersion` can't compare prefixed replacements** — the duplicate-edit merge at `transaction.go:397` feeds it `~1.3.0` vs `~1.4.0`, parsing fails, the left argument wins, and a two-advisories-on-one-declaration transaction aborts as "ineffective" where it could succeed.
4. **Yarn 2.x is classified as Berry** but `--mode=update-lockfile` exists only in Yarn 3+ — fail-safe failure; either detect the 2.x major or note it as unsupported.
5. **Fix-failure reporting** is still primary-format-JSON-only (`-o json=file`/markdown/SARIF specs get nothing) and still reuses `ResolutionFailureError`, conflating fix failure with dependency-resolution failure in exit semantics.
6. **Yarn/bun workspace members without their own lockfile still default to npm** in `detectManager`, so member-level `resolutions` are silently skipped (no ancestor walk).
7. Smaller: diff/explain markdown still emit a headed-but-empty `## Remediation` section (scan got the `Omit` treatment); `Base.Remediation` can now leak into the head-oriented diff report via the `reportOptionsFromPipelineResults` fallback since contexts are conditionally nil; an explicit `diff --auditors +override-hygiene` is silently overridden by the new exclude; the populated diff-TUI Remediation tab and ctx-cancellation mid-transaction remain untested; and the variadic `remediationContexts` and trailing-`:` override-evidence nits stand.

### Scope guardrails before this lands

Beyond the bug list, I'd codify the feature's boundary as an explicit contract rather than an implementation detail, so it stays stable as the surface grows:

- **State the `--fix` scope as a design principle, not a current limitation.** `docs/REMEDIATION.md` should carry a non-goals section: automatic application is — by design — deterministic, semver-compatible, direct, registry-declared, single-project, and provable from local evidence. Anything requiring judgment (major/breaking upgrades, transitive overrides, migration rewrites, cross-repo coordination) is out of scope for local execution, permanently classified as manual. Freezing that line now prevents scope creep from turning a provably-safe mechanism into a heuristic one, and it gives users a stable expectation of exactly what the CLI will and will not touch.
- **Version the remediation plan as a machine-readable contract.** The canonical context is already the single source every surface projects from; give the plan document an explicit schema version (as `mcp/1` did) and document it as a stable interface. That lets any external tooling consume plans and act on the manual/unavailable classes without coupling to CLI internals — the plan format, not the fix executor, is the extension point.
- **Keep the derivation evidence-only.** Detector hints are already demoted to evidence with core-owned safety policy; hold that rule for any future hint source so authority to mutate never migrates outward from the transaction core.

None of this blocks on new code beyond the doc section — it's mostly writing down boundaries the implementation already enforces.

internal/registry/builder.go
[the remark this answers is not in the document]
bomly-guy: This was addressed in 655f47f and tightened in f11c1d9: `internal/engine/diff/diff.go` excludes `override-hygiene` from both focused-graph audit runs (`withoutFocusedOverrideAudit`), and `internal/cli/diff_cmd.go` now rejects an explicit `--auditors +override-hygiene` on diff with an actionable error instead of silently overriding it. Override evidence still reaches diff output via `DeriveRemediation` against the full head graph. Covered by `internal/engine/diff/diff_test.go` and the diff_cmd validation.

bomly-guy:
Re-verified at `f11c1d9`: every item from my previous comment's **Still open** list is now addressed — CodeQL bounds guard (both alerts auto-closed as fixed), the satisfying-range no-op (solved via pin → regenerate → restore-range two-phase apply, with failed transactions no longer reporting changed files), prefix-aware `higherDeclarationVersion`, Yarn 2 explicitly demoted to manual at derivation with Yarn 3+ using `--mode=update-lockfile`, fix failures now reported across all requested formats with dedicated exit code 6, ancestor-walking manager detection for workspace members, diff/explain markdown `Omit`, base-side derivation removed with head remediation pinned, `diff --auditors +override-hygiene` now an actionable error, populated diff-TUI Remediation tests, a ctx-cancellation rollback test, and the variadic/trailing-colon cleanups. The scope-contract documentation also landed (plan `schema_version`, design-principle non-goals, evidence-only plugin boundary). All affected package tests pass locally; resolved the two CodeQL threads and the three CodeRabbit threads (the builder.go one was stale — see inline reply). No open findings remain from my reviews.

internal/engine/package_remediation.go
```diff
@@ -0,0 +1,175 @@
...
+		return vulnerabilityRemediationEvidence{unavailable: explicitlyUnavailable}
+	}
+	if explicitlyUnavailable {
+		return vulnerabilityRemediationEvidence{
+			hasFixEvidence: true,
+			contradictory:  true,
+		}
+	}
+
+	version, comparable := lowestComparableVersion(values)
```
bomly-guy: **Correctness: `recommended_version` can be below the installed version (effectively a downgrade recommendation).**

`derivePackageRemediation` never sees the package's current version. Per-vulnerability it takes the *lowest* comparable version from `FixAvailable`/`FixedVersions`, but multi-branch advisories list a fix per release line (e.g. minimist `CVE-2021-44906` → fixed in `0.2.4` and `1.2.6`). For `minimist@1.2.5` this picks `0.2.4`, and the cross-vulnerability `highest` step can still end below `1.2.5` — so a `status: complete` package can carry `recommended_version: 0.2.4`, which `groupRecommendedVersion` then surfaces to MCP agents as the version to move to.

`derivePackageRemediations` already has `pkg` in hand, so the installed version is available. Suggestion: when the package version parses, restrict candidates to fixed versions `> current` (falling back to `partial` when none qualify or the comparison isn't possible). That keeps the conservative contract — better to say `partial` than recommend a downgrade.

(The pre-existing OSV `FixedIn` extraction has a milder form of the same branch-arbitrariness — it returns the first `fixed` event — but this PR widens the exposure by folding in the full `FixedVersions` list and promoting the result to a package-level recommendation.)

internal/mcp/compact_scan.go
```diff
@@ -50,7 +51,7 @@ func BuildCompactScan(run ScanRunResult) CompactScanResponse {
 		AuditRan:           run.AuditRan,
 	}
 
-	if !run.AuditRan {
+	if !run.AuditRan && !run.EnrichRan {
```
bomly-guy, line 54: **Minor MCP regression: an enriched scan of a clean project now returns neither packages nor remediations.**

Before this change, every non-audited scan returned the capped package inventory. Now `enrich=true, audit=false` skips the inventory branch, and when the project has zero vulnerable packages `buildRemediations` produces nothing — the agent gets only summary counts, where it previously got the package list. An agent that ran an enriched scan to *both* inventory and check a project loses the inventory exactly when the news is good.

Suggestion: fall back to the inventory when enrichment produced no remediation groups (or always include it for non-audited runs as before). Cheap to do and it keeps the response non-empty on clean projects.

internal/mcp/tool_scan.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func registerScanTool(s *server.MCPServer, mcpCtx Context) {
 	tool := mcplib.NewTool("bomly_scan",
-		mcplib.WithDescription("Scan a project for dependencies, vulnerabilities, and policy findings. Returns a compact, remediation-focused JSON summary sized for tool-result limits: vulnerable packages grouped by the concrete fix that closes them (which direct dependency to bump, in which manifest, to which version), plus coverage counts for everything omitted. For a security review pass audit=true and enrich=true. Drill into one package's full advisory detail with bomly_explain; for the complete scan document run `bomly scan --format json -o <file>` via the CLI instead of this tool."),
+		mcplib.WithDescription("Scan a project for dependencies and vulnerabilities. With enrich=true, the compact response groups vulnerable packages by the change that may address them. With audit=true, it also includes policy results. Use bomly_explain for full details about one package, or the CLI JSON format for the complete scan document."),
```
bomly-guy: **Agent-guidance regression: the rewritten tool descriptions drop most of the steering an agent uses to resolve vulnerabilities.**

The payload still carries all of it (verified: `action` taxonomy, `override_advice`, per-fix `fixed_in`, KEV/severity/EPSS ranking are unchanged), but the descriptions no longer *tell the agent* any of that. Lost across `tool_scan.go`/`tool_explain.go`/`tool_diff.go`:

- the action vocabulary (`direct-bump`, `transitive-override`, `lockfile-refresh`, `no-fix-upstream`, `policy-review`) and the existence of override advice for transitive cases;
- the ranking contract (known-exploited first, then severity/EPSS/fixability);
- the scan-first-then-explain-each-package workflow hint (explain's "drill-down companion" framing is gone);
- for a security review, that `enrich`+`audit` together is the intended combination.

Tool descriptions are the primary steering surface for MCP clients — an agent that doesn't know `override_advice` exists won't look for it. The old descriptions were long, but that length was doing work. Suggest restoring at least the action taxonomy + override-advice mention and the scan→explain workflow hint. Same applies to `docs/MCP.md`, whose `remediations` bullet lost the action/override-advice enumeration.

Nit: the new explain description says dependency paths come "with enrich=true" — paths are returned without enrich too.

bomly-guy (commented):
## Review: package remediation enrichment context

Reviewed with a focus on MCP regressions (does an agent still get everything it needs to resolve vulnerabilities?) and overall usefulness.

### Verdict

**The PR is useful and I'd support merging it after the inline points are addressed.** The package-level `remediation` summary is honest derived data (no network, no manifest inspection, engine-owned and overwritten after matching), and the biggest win is real: MCP agents now get remediation groups from `enrich` alone instead of having to know to also pass `audit`. Placement is disciplined — JSON on `packages[]`, explain only on the focused dependency, SARIF/text/Markdown untouched, schema versions unchanged — and the test coverage is genuinely thorough (order-independence, idempotence, overwrite semantics, projection-mutation checks, size budget, protocol-v1 JSON round-trip).

### MCP regression check — payload: no loss; steering text: partial loss

I traced every field an agent previously relied on:

- Per-fix `fixed_in`, `override_advice`, the `action` taxonomy, KEV/severity/EPSS ranking, truncation counters — **all still present and unchanged** in the compact shapes.
- `remediationFindings` appends unmatched audit findings after the enrichment join, so license/package policy findings and the informational bucket are **not** dropped.
- The stricter `groupRecommendedVersion` (single affected PURL + `complete` status) is an improvement, not a regression: the old `max(fixed_in)` across a multi-package group was semantically wrong for the target dependency anyway, and the "no single safe version was computed" prose still fires as a fallback.

The regressions I did find are in the *guidance* layer, not the payload — see the inline comments on `tool_scan.go` and `compact_scan.go`, plus a correctness issue in the version derivation (`package_remediation.go`).

### Minor notes (no action strictly required)

- `remediationFindings` is O(packages × vulns × auditFindings). Fine under MCP caps, but indexing audit findings by `PackageRef` first would remove the quadratic scan on large audited scans.
- If audit ever emits two findings for the same vulnerability, only the first is overlaid; the second passes through as a leftover and would appear twice in groups. Probably unreachable today — worth a defensive note.
- A single contradictory vulnerability (`wont-fix` + a fixed version) collapses the whole package to `unknown`, even when the other vulns have complete evidence. `partial` arguably describes that state better; the current choice is defensible but worth a doc sentence.
- Enrich-only remediation means every enriched vulnerability is now a group candidate (previously only policy-failing findings were). The 40-group/15-per-group caps plus truncation counters handle this, and `TestCompactScanSizeStaysUnderBudget` covers it — just noting the cap will be hit more often on legacy-heavy projects.
- Java/Python enrichment goldens weren't regenerated locally (as the PR notes) — remember to dispatch the Update Smoke Goldens workflow after merge so `remediation` blocks land in those goldens before the nightly smoke run flags drift.

docs/detectors/ecosystems/rust/README.md
```diff
@@ -4,14 +4,15 @@
 
 Package managers Bomly recognizes in the `rust` ecosystem:
 
-| Package manager | Detector chain | Evidence patterns | Install-first support |
-| --- | --- | --- | --- |
-| [`cargo`](cargo.md) | `cargo-detector`, `syft-detector` | `Cargo.lock`, `Cargo.toml` | Yes |
+| Package manager | Detector chain | Evidence patterns | Remediation hints | Install-first support |
+| --- | --- | --- | --- | --- |
+| [`cargo`](cargo.md) | `cargo-detector`, `syft-detector` | `Cargo.lock`, `Cargo.toml` | `direct-bump`, `lockfile-refresh` | Yes |
```
bomly-guy, line 9: I don't know if Remediation hints such as `direct-bump`, `lockfile-refresh` are clear enough for user to understand. Perhaps we can use footnotes whenever values are added to docs to explain what they mean.

internal/detectors/remediation.go
```diff
@@ -0,0 +1,173 @@
...
+		return []sdk.RemediationAction{
+			sdk.RemediationActionDirectBump,
+			sdk.RemediationActionLockfileRefresh,
+		}
+	default:
+		return nil
+	}
+}
+
+func overrideAdvice(manager sdk.PackageManager, name, version, manifestPath string) string {
```
bomly-guy: I think it would be better if this belonged to individual detectors supporting override advice and became part of the SDK contract too. Ideally, defined closely or as part of the remediation hints since they are most likely to be paired with transitive override.

internal/mcp/remediation.go
```diff
@@ -172,93 +264,117 @@ func buildCompactFinding(f sdk.Finding, vuln *sdk.Vulnerability, in remediationI
...
-	case ActionPolicyReview:
-		group.Recommendation = "Policy finding: requires review, not fixed by a version upgrade."
+		group.Recommendation = "Review the package and owning manifest; available evidence does not support a concrete change."
 	}
 }
 
+// remediationFindings joins enriched vulnerabilities with optional audit
+// findings. Enrichment supplies complete vulnerability coverage; audit
+// findings overlay policy fields without filtering out unaudited advisories.
+func remediationFindings(registry *sdk.PackageRegistry, auditFindings []sdk.Finding) []sdk.Finding {
```
bomly-guy: **[major] Audit dispositions are bypassed: vulnerabilities the user explicitly allowed (or that fall below `--fail-on`) resurface as actionable remediation groups.**

`remediationFindings` synthesizes a finding for every vulnerability on every registry package, then overlays audit findings. But the vulnerability auditor emits **no finding at all** for vulns matched by `allow_vulnerability_ids` or failing `MatchesConstraints` (`internal/auditors/vulnerability/auditor.go:73-74` — it `continue`s). So with `audit=true`, an allowed CVE gets a synthetic `Auditor: "enrichment"` finding with empty `policy_status`, lands in `findingsByPackage`, and becomes a ranked, unmarked direct-bump group — indistinguishable from a failing finding. An MCP agent will happily "fix" the exact vulnerability the user suppressed. This affects scan, diff, and explain (all three wire in `remediationFindings`).

The synthesis needs to consult the audit configuration (or the auditor needs to emit suppressed findings with a status the projection can filter on), and warn/suppressed dispositions should keep routing to informational as before. No test covers allowed/suppressed vulns in MCP.

internal/remediation/derive.go
```diff
@@ -0,0 +1,806 @@
...
+		pkg.Remediation = derivePackageRemediation(pkg.Version, pkg.Vulnerabilities)
+	}
+}
+
+func derivePackageRemediation(currentVersion string, vulnerabilities []sdk.Vulnerability) *sdk.PackageRemediation {
+	if len(vulnerabilities) == 0 {
+		return nil
+	}
+
+	current, _ := semver.NewVersion(strings.TrimSpace(currentVersion))
```
bomly-guy: **[major] An unparseable installed version silently disables the "newer than installed" guard — a previously-flagged bug class re-entering.**

`current, _ := semver.NewVersion(...)` swallows the parse error, and the `versionsNewerThan` filter only runs `if current != nil` (line 129), while `selectComparableVersion` accepts a single (or repeated-identical) unparseable string as "comparable". Verified empirically: `derivePackageRemediation("1:2.0", [{FixedIn:"1.5.0"}])` → `Complete/1.5.0` (a recommendation *below* the installed version); `("2:1.2.3-1ubuntu1", ...)` → `Complete` with a distro-epoch string; `("", [{FixedIn:"see advisory"}])` → `Complete/"see advisory"` — which then flows into override advice as `pin foo to see advisory`.

Epoch/distro versions are routine in container scans. When the installed version doesn't parse, the honest status is `partial`/`unknown`, not `complete` — the contract says "a comparable recommended version must be newer than the installed version", and this is the one path where that isn't enforced. The existing test only covers unparseable *fix* versions with a parseable current.

internal/remediation/derive.go
```diff
@@ -0,0 +1,806 @@
...
+	fixed := make([]string, 0, len(vulnerability.FixedVersions))
+	for _, version := range vulnerability.FixedVersions {
+		if value := strings.TrimSpace(version); value != "" {
+			fixed = append(fixed, value)
+		}
+	}
+	return uniqueSortedVersions(fixed)
+}
+
+func versionsNewerThan(values []string, current *semver.Version) ([]string, bool) {
```
bomly-guy, line 246: **[major] No prerelease guard anywhere in the derivation — `1.0.0` with `FixedIn: "2.0.0-rc.1"` yields `Status: complete, RecommendedVersion: "2.0.0-rc.1"` (verified empirically).**

`versionsNewerThan` uses plain `GreaterThan`. OSV advisories do list rc/beta as first-fixed versions, so this pins users to prereleases via direct-bump suggestions and bakes them into manager advice (`add "overrides": {"x": "2.0.0-rc.1"}`). This exact class was flagged (and fixed) in the previous iteration of this PR. Either skip prerelease candidates when a stable fix exists, or demote to `partial` with a limitation note when only a prerelease fixes the advisory. No test exercises prerelease.

internal/remediation/derive.go
```diff
@@ -0,0 +1,806 @@
...
+		for _, strategy := range hint.Strategies {
+			if _, ok := advertised[strategy.Action]; !ok {
+				rejected = append(rejected, fmt.Sprintf("ignored unadvertised remediation action %q for dependency %q", strategy.Action, dependencyRef))
+				continue
+			}
+			switch strategy.Action {
+			case sdk.RemediationActionDirectBump,
+				sdk.RemediationActionTransitiveOverride,
+				sdk.RemediationActionLockfileRefresh:
+				strategies[strategy.Action] = strings.TrimSpace(strategy.Advice)
```
bomly-guy: **[major] Plugin-supplied advice and diagnostics are neither bounded nor sanitized before reaching the terminal and size-capped responses.**

`strings.TrimSpace(strategy.Advice)` is the only processing — no length cap, no control-character stripping. Diagnostics are forwarded verbatim into pipeline warnings (line 272-276 → `pipeline.go:306-321`), and `internal/tui/utils.go:374-377` renders `OverrideAdvice` directly into the terminal. An external detector plugin can emit ANSI escape sequences (terminal spoofing in a security-sensitive CLI) or megabyte-scale strings / unbounded diagnostic floods that also blow through MCP's compact-response size discipline (`suggestion.OverrideAdvice` flows uncapped into `RemediationGroup.OverrideAdvice`/`Recommendation`).

`validateHints` is the right chokepoint: cap advice length (a sentence or two), strip control characters, and bound the diagnostic count per detector there, so every downstream surface inherits the guarantee.

test/smoke/helpers_test.go
```diff
@@ -479,6 +480,27 @@ func scrubAdvisoryFeedMap(m map[string]any) {
...
 }
 
+// normalizePackageRemediationVersions keeps the derived status in smoke
+// goldens while scrubbing the recommended version sourced from live advisory
+// feeds. The recommendation can change as those feeds publish new fixes.
+func normalizePackageRemediationVersions(node any) {
+	switch v := node.(type) {
+	case map[string]any:
+		if remediation, ok := v["remediation"].(map[string]any); ok {
+			if _, exists := remediation["recommended_version"]; exists {
```
bomly-guy, line 490: **[major] Two stale goldens will fail the next nightly smoke run, and this normalizer misses the version literal inside `override_advice`.**

(a) All four reachability cases run `--enrich` against live feeds, and the pipeline now emits `remediation` for every package with vulnerabilities — but only the go and npm goldens were regenerated. `scan-python-pip-reachability.golden.json` (17 packages, 41 vulns) and `scan-java-maven-reachability.golden.json` (64 packages, 94 vulns) contain **zero** `remediation` keys and will fail comparison. Regenerate both (Update Smoke Goldens dispatch covers the missing-Java/Python host limitation noted in the PR body).

(b) `normalizePackageRemediationVersions` scrubs `recommended_version`, but the same live version is embedded verbatim in `suggestions[].override_advice` — the npm golden carries e.g. `add "overrides": {"ansi-regex": "3.0.1"}...` in all 17 transitive-override entries. First OSV update to any of those packages breaks the golden — exactly the drift this normalizer exists to prevent. Scrub the version inside the advice string (or normalize the whole advice value).

internal/mcp/remediation.go
```diff
@@ -30,61 +33,65 @@ type remediationOutput struct {
...
+			packageRemediation = in.FocusedRemediation
+		}
+		if packageRemediation == nil || len(packageRemediation.Suggestions) == 0 {
+			for _, fix := range fixes {
+				appendInformational(&informational, fix, trunc)
+			}
+			delete(findingsByPackage, pkg.PURL)
+			continue
+		}
+		for _, suggestion := range packageRemediation.Suggestions {
```
bomly-guy, line 77: **[minor] Truncation counters overcount at both caps.**

At the group cap, the check is per suggestion inside the per-package loop and each capped suggestion adds `len(fixes)` for the *same* finding set — a package with two capped suggestions counts its findings twice, and if its first suggestion made it into `groups`, the second still adds findings that are in fact visible. At the per-group cap, a package with 2 suggestions and 20 findings reports `OmittedFindings += 5` twice while only 5 distinct findings are invisible (both groups keep the same sorted first 15). `TestBuildRemediationsTruncatesWithCounters` only asserts non-zero; neither exactness path is tested. Count distinct omitted findings per package (set semantics), not per suggestion.

internal/mcp/remediation.go
```diff
@@ -30,61 +33,65 @@ type remediationOutput struct {
...
+		findingsByPackage[f.PackageRef] = append(findingsByPackage[f.PackageRef], compact)
+	}
 
-		action := remediationAction(f, vuln, compact, ancestor, ancestor.packageManager)
-		key := remediationGroupKey(action, ancestor, compact)
-		group, ok := groups[key]
-		if !ok {
+	var groups []RemediationGroup
+	for _, pkg := range packagesWithRemediation(in.Registry) {
+		fixes := findingsByPackage[pkg.PURL]
```
bomly-guy, line 61: **[minor] Every suggestion for a package carries *all* of the package's findings, inflating `Summary.Actionable`, `FindingsBySeverity`, and response size.**

Canonical suggestions are occurrence-scoped (one per manifest/target/action — a multi-module workspace package legitimately has several), but `fixes` is passed unchanged into each `canonicalRemediationGroup`, and `compact_scan.go:72-77` sums `len(group.Fixes)` per group — so a package present in two manifests double-counts every advisory and duplicates the same `CompactFinding` payloads across up to 40 groups. The old projection placed each finding in exactly one group. Scope fixes to the findings actually relevant to the suggestion's occurrence (via `DependencyRefs`), or count distinct findings in the summary.

internal/detectors/remediation.go
```diff
@@ -0,0 +1,173 @@
...
+				ManifestPath:  entry.Manifest.Path,
+			}
+			for _, action := range remediationActionOrder {
+				if _, ok := actions[action]; !ok {
+					continue
+				}
+				strategy := sdk.RemediationStrategyHint{Action: action}
+				if action == sdk.RemediationActionTransitiveOverride {
+					strategy.Advice = overrideAdvice(
+						manager,
```
bomly-guy: **[minor] Lockfile-refresh suggestions lost their concrete command advice — a capability regression vs the old MCP table.**

Advice is attached only for `RemediationActionTransitiveOverride`; lockfile-refresh returns `""` (asserted empty in `derive_test.go:361`), so MCP falls back to generic "Refresh the resolved version in go.mod." The deleted `TestTransitiveGoModuleGetsLockfileRefresh` previously guaranteed a copy/paste `go get <pkg>@v<ver> && go mod tidy` command for gomod/cargo/bun transitive cases — that end-to-end capability is gone and no MCP test exercises `ActionLockfileRefresh` anymore. Since this file already owns per-manager advice templates, restoring the command-form advice for lockfile-refresh (gomod/cargo/bun) is cheap and directly serves the agent-consumer use case.

internal/mcp/remediation.go
```diff
@@ -30,61 +33,65 @@ type remediationOutput struct {
...
-		group.Fixes = append(group.Fixes, compact)
+		delete(findingsByPackage, pkg.PURL)
 	}
-
-	out := make([]RemediationGroup, 0, len(groups))
-	for _, key := range groupKeys {
-		group := groups[key]
-		finalizeGroup(group)
-		out = append(out, *group)
+	for _, fixes := range findingsByPackage {
```
bomly-guy: **[minor] Leftover `findingsByPackage` is a map iterated in random order before the `maxInformational` cap — which findings survive vs count as omitted varies run to run once the cap binds.**

Everything else in the projection is deterministic (PURL-sorted registry, sorted suggestion keys); sort these keys too so MCP responses stay reproducible.

internal/remediation/derive.go
```diff
@@ -0,0 +1,806 @@
...
+		return nil
+	}
+	clone := sdk.NewPackageRegistry()
+	for _, pkg := range registry.All() {
+		clone.Add(pkg.Clone())
+	}
+	return clone
+}
+
+func cloneDetectionResult(result sdk.DetectionResult) sdk.DetectionResult {
```
bomly-guy, line 853: **[minor] Clone isolation gap: `clone := result` copies `SubprojectInfo` by value, but `sdk.Subproject.DetectedPackageManagers`/`PlannedDetectors` slices share backing arrays with the engine's `ResolveResults`.**

An in-process provider mutating `req.Detection.SubprojectInfo.DetectedPackageManagers[0]` corrupts scan output. Graphs, packages, and the registry are properly deep-cloned (tests prove it), and external plugins are isolated by gRPC serialization — hence minor — but the in-process contract says "cloned detection data" and this field isn't. The existing mutation test only touches `Manifest.Path` and the registry.

internal/remediation/derive.go
```diff
@@ -0,0 +1,806 @@
...
+			validated[idx].detectorName = detection.DetectorName
+			manifestPath, dependencyRef, ok := selectedOccurrence(
+				in.ProjectPath,
+				in.Manifests,
+				detection.DetectorName,
+				validated[idx].manifestPath,
+				validated[idx].dependencyRef,
+				validated[idx].sourceDependencyRef,
+			)
+			if !ok {
```
bomly-guy, line 352: **[minor] A validated hint that fails consolidated-manifest resolution is silently dropped — the only rejection path with no `Warning`.**

`selectedManifestForDependency` requires the consolidated graph to contain a node keyed by the detection-time ID or canonical PURL, and consolidation can rebase/dedup node IDs (that exact class caused the manifest-rebase fix in the recursive-discovery work). When it misses, the occurrence falls back to manual-review with zero observability. Emit a warning like every other rejection, and add a consolidation-rebase test for the hint path.

internal/remediation/derive.go
```diff
@@ -0,0 +1,806 @@
...
+		}
+		if len(parents) == 0 {
+			root, _ := graph.Node(state.nodeID)
+			if executableRoot(root) && len(state.path) >= 2 {
+				shortest = distance
+				candidates = append(candidates, state)
+			}
+			continue
+		}
+		for _, parent := range parents {
```
bomly-guy: **[minor] `inferredPlacement` enumerates *all* equal-length shortest paths (the `previous < nextDistance` visited check deliberately re-enqueues equal-distance states, each materializing a full path copy) — worst case exponential.**

A chain of k same-depth diamonds in the dependents graph yields 2^k stored paths before the lexicographic tie-break. Large npm graphs with heavy same-depth fan-in can hit real CPU/memory blowup in the enrichment tail. The tie-break itself is correctly deterministic — tracking a single lexicographically-best predecessor per node gives the same answer in linear time.

bomly-guy (commented):
Full review of the rewritten approach (head `f68e823`). **The architectural pivot is sound and I endorse it**: remediation as a pure, read-only projection over enrichment evidence — with all decision authority centralized in `internal/remediation`, detectors demoted to capability-scoped advice providers, and no write path at all — is a much better foundation than the transactional design. What verified clean: the enrichment-tail ordering matches the stated contract; provider failures degrade to manual-review and never abort; synthetic roots are excluded from parent evidence with tests; output ordering is deterministic by construction (sorted nodes, sorted suggestion keys, full tie-breaks); SARIF is byte-identical to main with a test pinning that; CLI JSON stays additive on schema 1.0; protocol-v1 is genuinely additive (unadvertised plugins never called); package boundaries hold (`internal/detectors/remediation.go` imports only sdk; no new deps); `make generate` is drift-free; docs (MODELS, ARCHITECTURE decision log, PLUGINS, TUI, SCHEMAS) accurately describe the implementation; and there are zero stale references to the abandoned `--fix`/overrides design. All tests pass and CI is green.

That said, the version-derivation core re-admits two previously-flagged bug classes, the MCP projection has one significant audit-semantics bypass, and two smoke goldens are stale (CI-breaking) — inline comments cover these.

### Items without inline anchors

- **`ActionPolicyReview` removed from the mcp/1 vocabulary** (`types_compact.go`): failing license/policy findings are demoted from actionable groups to informational, and `Summary.Actionable` no longer counts them. The struct contract is intact, but the PR body claims "preserving its existing ranking, caps, policy overlays" and "additive only" — this is a behavioral vocabulary change for mcp/1 consumers keying on policy-review groups. It looks intentional (tool descriptions were updated consistently); state it explicitly in the PR body/docs rather than implying pure additivity.
- **`compact_explain.go:143` — explain filtering matches only `f.PackageRef == purl`**; a finding carrying only `DependencyRefs` (validate.go doesn't require `PackageRef`) is silently dropped with no truncation counter. Common paths always set `PackageRef`, so minor — but this was flagged in a prior round and remains.
- **Dead computation**: both `buildCompactFinding` call sites discard the `ancestorTarget` return (`internal/mcp/remediation.go:47`, `compact_diff.go:121`) — the struct and its per-finding manifest scan are computed and thrown away. Either use it or remove it.
- **TUI diff detail pane passes no dependency ref** (`internal/tui/diff.go:1364/1375/1384`) while scan filters to the selected occurrence — in a multi-module diff, one occurrence's details list override advice for unrelated manifests. `change.Package.ID` is available on the adjacent line.
- **`bomly plugin list` shows no remediation capability column** — only `plugin info <id>` does; the PR body's "detector listings" slightly overstates. Fine either way, but align the text or add the column.
- **Latent: `sdk.Package.MergeFrom` never carries `Remediation`** — currently unreachable (derive runs after all merges) but any future post-derive merge silently drops it; worth a one-line copy + test.
- **Build-metadata ties are order-dependent** (`derive.go:234-238`): Masterminds ignores build metadata, so `1.2.3+a` vs `1.2.3+b` keeps whichever arrives first; the order-independence test only covers distinct versions.
- **Contract deviation (nit)**: a non-registry occurrence of a wontfix package gets `no-fix-upstream` (`derive.go:637-644` checks Unavailable before registry eligibility), while the stated rule says non-registry occurrences are manual. Arguably better behavior — but then the docs/contract should say so, and a test should pin it.
- **Hint plumbing nits**: duplicate hints for one occurrence are last-wins with no warning (`derive.go:532-536`); after ctx cancellation `collectHints` still iterates remaining detectors emitting one warning each (`derive.go:263-270`).
- **Observability**: `Derive` has no INFO stage boundary or per-detector DEBUG counts, and `externalDetector.RemediationHints` (`internal/plugin/registry.go:144-162`) launches a plugin subprocess with no DEBUG line — CLAUDE.md requires subprocess invocations be reproducible at `-vv`.
- **`Descriptor()` now deep-clones on every call** (`internal/registry/builder.go:130-132`), including hot planning loops, and the sdk Clone converts nil slices to empty non-nil ones — consider cloning once at registration or only on mutation-risk paths.
- **Nit**: `remediationActionLabel`/status capitalization byte-slice `value[:1]` (`internal/tui/utils.go:354/389`) — safe for today's internal ASCII enums, would garble a non-ASCII first rune.

docs/detectors/ecosystems/rust/README.md
bomly-guy: Added plain-language explanations beside the generated ecosystem tables and footnotes on every individual package-manager page that advertises these values. The same simple definitions now appear in the MCP guide. Implemented in 5d01512 and c1f1f7a.

PR #297, opened 2026-07-23
Add portable finding baselines

bomly-guy:
## Summary

- add a versioned, package-specific finding baseline format with automatic project discovery for audited local projects
- resolve finding policy status as the final part of the existing audit stage
- add `bomly baseline create|update|prune|inspect` lifecycle commands
- expose `--baseline auto|none|path` consistently through scan, diff, explain, configuration, and MCP
- keep accepted findings visible through existing output models

## Architecture

Baselines are an audit policy input, not a separate pipeline stage or output domain. Auditors still produce complete reference-style findings. The shared SDK now owns the neutral `FindingPolicyResolver` contract and canonical status ranking; resolvers may only change policy status and cannot remove or rewrite evidence. The least suppressive valid result wins when multiple resolvers contribute.

Baseline identity is portable across projects. Entries use the full package PURL, finding kind, auditor, and advisory aliases or a stable rule ID. Dependency occurrences, manifest paths, and project identity are intentionally excluded.

User-facing surfaces call the concept **policy status**: fail, warning, or accepted. Existing SDK and structured-output names (`Finding.Disposition` and `disposition`) remain unchanged for compatibility. The new baseline document uses `policy_status`.

## Safety and behavior

- automatic baseline discovery is disabled for `--url` targets; an explicit trusted absolute path is required
- baselines load only for audited commands, so malformed automatic policy cannot break inventory-only scans
- loaded policy path and entry count are logged at INFO; missing or ignored automatic policy is visible at DEBUG
- severity, policy status, and reachability are escalation bounds; safer reachability transitions remain accepted
- unknown resolver statuses are ignored and cannot weaken enforcement
- create refuses overwrite
- update retains historical accepted entries
- prune removes only absent accepted entries and never accepts new findings
- writes use atomic replacement and reject symbolic-link destinations
- lifecycle mutation is limited to writable local project directories
- degraded detector, matcher, analyzer, or auditor coverage prevents mutation
- baseline consumption adds no network access

## Output impact

Finding collections and integration formats remain unchanged. Accepted findings stay present with `disposition: suppressed`; failure counts exclude them. Stable `rule_id` is now included in JSON, SARIF properties, and compact MCP findings. SARIF emits first-class external suppressions, and the TUI reports warnings and accepted findings separately. Text and Markdown continue using their existing finding projections. Bomly Guard continues consuming ordinary diff output and exit status.

The shared policy helper also makes `warn_only` behavior consistent for explain's component audit path.

## Validation

- `make generate`
- `make lint`
- `make test`
- `make build`
- focused baseline, pipeline, diff, explain, output, SDK, CLI, MCP, and TUI tests
- end-to-end baseline smoke test against a pinned repository
- `git diff --check`
- generated documentation drift check

## Summary by CodeRabbit

- **New Features**
  - Added finding baselines (`baseline create`, `update`, `prune`, `inspect`) selectable via configuration, CLI flags, and MCP.
  - Baseline-matched findings are marked `suppressed` (non-gating) while remaining visible in results, including SARIF.
  - Added stable `rule_id` plus updated finding policy output using `policy_status`.
- **Bug Fixes**
  - Policy resolution now consistently preserves finding evidence and applies the least-suppressive applicable policy outcome.
  - Warn/non-gating and suppressed acceptance behavior is reflected consistently across outputs and the diff UI.
- **Documentation**
  - Updated user and architecture docs for baselines, `baseline` arguments, and the `policy_status`/`rule_id` output changes.


internal/engine/disposition.go
```diff
@@ -0,0 +1,64 @@
+package engine
```
bomly-guy: Wouldn't these new types and interface belong better to the SDK?

internal/baseline/baseline.go
```diff
@@ -0,0 +1,415 @@
...
+		return Document{}, fmt.Errorf("parse baseline %q: trailing JSON content", path)
+	}
+	if err := document.Validate(); err != nil {
+		return Document{}, fmt.Errorf("validate baseline %q: %w", path, err)
+	}
+	return document, nil
+}
+
+// ResolversForTarget discovers or loads the selected baseline for target.
+func ResolversForTarget(selection string, target sdk.ExecutionTarget) ([]engine.FindingDispositionResolver, error) {
```
bomly-guy: **Auto-discovery trusts the scanned tree — and applies silently.** `--baseline` defaults to `auto`, so `bomly scan --url <third-party repo> --enrich --audit` applies *that repo's* committed `.bomly/baseline.json`: the scanned project can suppress its own failing findings and flip the exit code to pass. That's the right default for Guard/own-project workflows, but for auditing someone else's code it's a silent policy input.

Compounding it, baseline discovery/loading emits no log at all — the only trace is the `suppressed` count inside the audit INFO log. Per the repo's logging checklist, this function (or `Load`) should log which file was applied and how many entries at INFO, and DEBUG the resolved path when nothing is found. I'd also add a line to `docs/BASELINES.md` recommending `--baseline none` when scanning untrusted repos.

internal/cli/opts/options.go
```diff
@@ -328,20 +331,29 @@ func (o *Options) PrepareForExecutionTarget(ctx context.Context, logger *zap.Log
 		return Options{}, err
 	}
 
+	resolvers, err := baseline.ResolversForTarget(resolved.Baseline, executionTarget)
```
bomly-guy: **A malformed auto-discovered baseline fails every command, including non-audit scans.** This runs unconditionally during preparation, and any parse/validation error (not just `ErrNotExist`) aborts even a plain `bomly scan` with no `--audit`. `docs/BASELINES.md` only promises "a malformed **explicitly selected** file fails before the audit runs."

Fail-fast may be intended, but for `auto` + no audit a warning-and-continue is arguably better — either way, please pin the intended behavior with a test and document it.

internal/baseline/baseline.go
```diff
@@ -0,0 +1,415 @@
...
+		for _, right := range actual.AdvisoryIDs {
+			if strings.EqualFold(left, right) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func stateCompatible(expected, actual Entry) bool {
```
bomly-guy, line 420: **Reachability bound uses equality, not ordering.** A baseline created without `--analyze` records `unknown`; a later `--analyze` scan reporting `unreachable` — a strictly *safer* state — un-suppresses the finding. Only `reachable` should escape the bound; moving to `unreachable` shouldn't invalidate acceptance.

Related usability trap: `baseline create` requires `--enrich` but nothing nudges analyze-parity with the team's normal scan flags, which makes this mismatch easy to hit. Worth deciding now since it affects the baseline format's compatibility story.

internal/engine/disposition.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+		}
+		if matched {
+			out[idx].Disposition = selected.Disposition
+		}
+	}
+	return out
+}
+
+func dispositionRank(disposition sdk.FindingDisposition) int {
```
bomly-guy: **Duplicate `dispositionRank` with different scales.** This one uses suppressed=1…default=4 while `internal/baseline/baseline.go` has its own with suppressed=0…default=2. They serve different comparisons today, but the duplication invites drift — consider exporting one ranking from `engine` (or `sdk`) per the repo's "use canonical shared types" convention.

sdk/vulnerability.go
```diff
@@ -531,16 +534,19 @@ func (v Vulnerability) IsExploitable() bool {
...
+	Kind        FindingKind        `json:"kind"`
+	Title       string             `json:"title,omitempty"`
+	Severity    SeverityLevel      `json:"severity,omitempty"`
+	Disposition FindingDisposition `json:"disposition,omitempty"`
+	Reasons     []string           `json:"reasons,omitempty"`
+	Source      string             `json:"source,omitempty"`
+	Auditor     string             `json:"auditor,omitempty"`
+	// RuleID is the stable auditor rule that produced the finding. Unlike ID,
+	// it must not contain package versions or project-specific occurrence data.
+	RuleID           string    `json:"rule_id,omitempty"`
```
bomly-guy, line 550: **`rule_id` isn't exposed in any output projection.** `output.AuditFinding` (JSON/SARIF/MCP compact) omits `RuleID`, so users can't hand-author non-vulnerability baseline entries from scan output — they must run `baseline create` or read source. Adding it to the JSON projection is additive and cheap (would need `make generate`).

internal/cli/render/diff_markdown.go
```diff
@@ -378,7 +378,7 @@ func findingIcon(status, disposition string) string {
 	if status == "resolved" {
 		return "✅"
 	}
-	if strings.EqualFold(disposition, "warn") {
+	if strings.EqualFold(disposition, "warn") || strings.EqualFold(disposition, "suppressed") {
```
bomly-guy: This markdown icon fix has no TUI counterpart: `internal/tui/diff.go`'s verdict switch (~line 2037) still counts only `fail`/`warn`, so suppressed introduced findings render as "PASS (warn-only) — N new finding(s), all Disposition=warn" and are missing from every count. See the review body for details.

internal/engine/pipeline.go
```diff
@@ -394,6 +406,14 @@ func (p *Pipeline) auditComponent(ctx context.Context, g *sdk.Graph, registry *s
 	}
 	result, err := p.engine.Audit(ctx, auditReq)
 	result.Findings = DeduplicateFindings(result.Findings)
+	if req.WarnOnly {
```
bomly-guy: Unmentioned behavior change: this adds `warn_only` rewriting to `auditComponent` (explain's per-component audit path), which previously didn't apply it. Sensible fix, but it's silent in the PR description and has no dedicated test — worth a line in the description and a small explain-path warn-only assertion.

bomly-guy (commented):
Solid, well-architected feature — the suppression semantics, validation, and portability model are carefully designed, and the wiring follows the repo's package boundaries closely (resolver interface in `engine`, implementation in `internal/baseline`, no `engine → baseline` import). I checked out the branch and verified `go build ./...` plus the `baseline`, `engine`, `output`, `sdk`, `mcp`, `config`, and baseline CLI tests all pass.

Inline comments cover the main findings. Two more that don't anchor to changed files:

**TUI diff verdict copy is now wrong for suppressed findings** (`internal/tui/diff.go:2037-2099`, not touched by this PR). The verdict switch counts only `fail`/`warn`; suppressed introduced findings fall into the `IntroducedTotal > 0` branch, which renders "PASS (warn-only) — N new finding(s), all Disposition=warn". With suppressed findings that label is inaccurate, and they're absent from every count. `render/diff_markdown.go` got the equivalent fix; the TUI didn't.

**Smoke tests are missing.** This is a user-visible feature touching scan/diff/explain output and exit codes, and the feature checklist calls smoke coverage the most-forgotten surface. No `test/smoke` changes and no slice registrations in `smoke.yml` / `update-smoke-goldens.yml`. At minimum: a scan case against a pinned repo with a committed baseline. Fine as an immediate follow-up if you prefer.

Minor:
- No unit tests for `ResolvePath` (auto/none/relative/container/SBOM cases), `ResolversForTarget` (missing-file tolerated for auto but required for explicit paths), or `WriteAtomic`'s refuse-overwrite path.
- `baseline inspect` resolves its default path against CWD while everything else resolves against the project root — inconsistent for out-of-root invocations.
- `Load` uses `strings.NewReader(string(data))`; `bytes.NewReader(data)` avoids a copy.
- SARIF maps suppressed → level `note`, which works, but SARIF has a first-class `suppressions[]` property; GitHub code scanning would then show alerts as *suppressed* rather than as new note-level alerts. Good follow-up.

PR #315, opened 2026-07-25
feat(matchers): declare supported ecosystems on built-in matchers

bomly-guy:
## Why

`sdk.MatcherDescriptor.SupportedEcosystems` has existed for a while, but no built-in matcher populated it — and `supportsEcosystem` treats an empty list as *all ecosystems*. That claim is wrong for two matchers, which silently skip packages they cannot map:

- **deps.dev** accepts exactly seven ecosystems in `depsDevSystem`; everything else is dropped without a word.
- **Grype** (builtin mode) maps ecosystems onto Syft package types; anything unmapped reaches Grype as `UnknownPkg` and matches nothing.

Because the docs generator only emits an Ecosystems row when the field is non-empty, no matcher doc page carried one, and `bomly plugins list` had no ecosystem column for matchers at all.

## What changed

| Matcher | Declares | Why |
|---|---|---|
| `depsdev-license-matcher` | 7 ecosystems | mirrors `depsDevSystem` |
| `grype` (builtin) | 19 ecosystems | mirrors `ecosystemToSyftType` |
| `grype` (external) | nil | grype CLI derives coverage from PURLs — genuinely unbounded |
| `osv` | nil | queries by PURL, so OSV.dev decides coverage |
| `scorecard` | nil | bounded by GitHub source-repo resolution, not ecosystem |
| auditors ×3 | nil | ecosystem-agnostic by construction |

Grype's declaration is split by build tag so external mode isn't under-claimed. The `nil` cases now carry comments explaining that nil is the *correct* answer rather than an unset field, so nobody "fixes" them later.

Two drift tests fail if a mapping and its declaration diverge:
- `TestDescriptorEcosystemsMatchSupportedSystems` (deps.dev)
- `TestSupportedEcosystemsMatchSyftTypeMapping` (grype, builtin only)

Docs regenerated via `make docs-components`, plus an `ECOSYSTEMS` column on the matcher table in `bomly plugins list` (renders `all` when unbounded).

## Worth a look before merging

While mapping Grype I found that **apk, dpkg, and rpm packages currently match nothing**: `graphPkgToGrypePkg` sets no `Distro` and `FindMatches` gets an empty package `Context`, but Grype's OS matchers are distro-namespace driven. They're declared here because the type mapping genuinely exists — but container OS packages won't produce vulnerabilities until the distro is plumbed through. That felt like a separate change requiring your judgment, so I left it alone.

## Testing

`go test ./...` — all 65 packages pass. Also verified `go build -tags bomly_external_grype ./...` and `go vet` for the external path.

## Summary by CodeRabbit

* **New Features**
  * Plugin listings now display matcher-supported ecosystems, including deduplicated and sorted values.
  * Matchers now clearly identify which ecosystems they support, including deps.dev and Grype coverage.

* **Documentation**
  * Updated matcher documentation with supported ecosystem lists.
  * Clarified how ecosystem support and applicability are determined across matcher and auditor types.
  * Improved plugin development guidance for declaring supported ecosystems.

* **Tests**
  * Added coverage to verify ecosystem declarations match supported integrations.

### DIFF
--- a/docs/matchers/depsdev-license-matcher.md
+++ b/docs/matchers/depsdev-license-matcher.md
@@ -13,6 +13,7 @@ Fetches package metadata from deps.dev to improve license coverage.
 | Cache behavior | Uses Bomly's matcher cache; cache failures are non-fatal. |
 | Output fields | license value, license source, matched package flag |
 | Tags | `license-enrichment`, `batch-http` |
+| Ecosystems | `npm`, `maven`, `go`, `python`, `dotnet`, `ruby`, `rust` |
 
 ## User notes
 
@@ -20,7 +21,7 @@ Run with `--enrich` when you want license metadata from deps.dev.
 
 ## What `depsdev-license-matcher` does
 
-`depsdev-license-matcher` fetches package metadata from [deps.dev](https://deps.dev) (Google's open package metadata service) and attaches license information to packages that the detector did not resolve a license for. deps.dev coverage is strongest for npm, Go, Maven, NuGet, PyPI, Cargo, and RubyGems.
+`depsdev-license-matcher` fetches package metadata from [deps.dev](https://deps.dev) (Google's open package metadata service) and attaches license information to packages that the detector did not resolve a license for. It covers the ecosystems listed in the table above; packages from any other ecosystem are skipped.
 
 ## When to use it
 
--- a/docs/matchers/grype.md
+++ b/docs/matchers/grype.md
@@ -12,6 +12,7 @@ Uses Grype vulnerability matching against the resolved dependency graph.
 | Uses network | No |
 | Cache behavior | Relies on Grype's local vulnerability database behavior. |
 | Output fields | vulnerability ID, severity, CVSS, fixed version, fix state, EPSS, CWE, known exploitation, risk score, references |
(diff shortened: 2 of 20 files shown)


bomly-guy:
Extended in `ad5992a`: OSV now declares its ecosystems too.

[osv.dev publishes what it covers](https://google.github.io/osv.dev/data/#covered-ecosystems) and the list is finite, so "nil means all" was wrong there as well — I'd originally left it nil reasoning that the PURL path means "OSV decides", but OSV deciding *is* a finite documented list. The descriptor now carries the intersection with Bomly's ecosystems (20 of them).

That also surfaced four gaps in `ecosystemToOSV`, which only runs for packages without a PURL and was silently skipping ecosystems osv.dev does cover: Scala (Maven Central), Elixir and Erlang (both Hex), OCaml (opam), and GitHub Actions.

C/C++ is deliberately excluded — OSV covers it through git and OSS-Fuzz rather than as a package ecosystem, so a Conan package has nothing to match against.

Companion PRs for the external plugins, which had the same unset-means-everything problem:
- bomly-dev/bomly-plugin-clearlydefined-matcher#6 — genuinely narrow (php, dpkg, swift, conda); it defers npm/Maven/PyPI to the built-in matchers
- bomly-dev/bomly-plugin-eol-matcher#6 — genuinely unbounded, so that one just documents why nil is correct

PR #321, opened 2026-07-25
fix(osv): emit spec purl types and make the name+ecosystem fallback reachable

### LINKED ISSUE #317: OSV: five declared ecosystems can never match (non-standard PURL types + dead name fallback)
## Summary

#315 declares the ecosystems osv.dev covers on the OSV matcher. Five of them cannot return results today, because Bomly emits a PURL type that OSV does not recognise and the name-based fallback that would have rescued them is unreachable.

Nothing errors — the queries just come back empty, so these ecosystems look clean rather than unchecked.

### 1. Non-standard PURL types

`Coordinates.CanonicalPURL` calls `PackageURLTypeForValues(i.Ecosystem, i.PackageManager, i.Type)`. The explicit switch is consulted for each value in order, and when nothing matches it falls through to returning the **first non-empty value verbatim** — which is the ecosystem. For most ecosystems the package manager rescues it (`cargo`, `nuget`, `pub`, `mix`), but where neither the ecosystem nor the manager has an explicit case, we emit a type that is not in the purl spec.

Verified by calling `PackageURLTypeForValues(ecosystem, manager)` directly:

| Ecosystem | Manager | Bomly emits | purl spec / OSV expects |
|---|---|---|---|
| erlang | rebar, otp | `pkg:erlang` | `pkg:hex` |
| haskell | cabal, stack | `pkg:haskell` | `pkg:hackage` |
| r | r-package | `pkg:r` | `pkg:cran` |
| ocaml | opam | `pkg:ocaml` | `pkg:opam` |
| dpkg | dpkg | `pkg:dpkg` | `pkg:deb` |

Everything else checked out: npm, maven, scala→maven, go→golang, python→pypi, dotnet→nuget, ruby→gem, rust→cargo, php→composer, dart→pub, swift→swift, swift/cocoapods→cocoapods, elixir→hex, github-actions→githubactions, apk, rpm, cpp→conan.

Note `ocaml`: the manager is literally `opam`, which *is* the right purl type — it loses only because the ecosystem is checked first and the fallback returns the first value rather than the most specific one.

### 2. The name+ecosystem fallback is dead code

`buildQuery` looks like it degrades gracefully:

```go
// Fall back to name + ecosystem + version
ecosystem := ecosystemToOSV(string(dep.Ecosystem))
```

but the caller never lets it happen:

```go
purl := sdk.CanonicalPackageURLFromDependency(dep)
if purl == "" {
    stats.skippedPackages++
    continue          // <- buildQuery is never reached with an empty purl
}
```

and `PackageURLTypeForValues` cannot return `""` anyway — its final line is `return "generic"`. So `ecosystemToOSV` is currently unreachable at runtime, and a wrong PURL type has no second chance.

## Impact

`erlang`, `haskell`, `r`, `ocaml`, and `dpkg` are declared as OSV-supported in #315 but return nothing. `dpkg` is the one that stings: Debian and Ubuntu are among OSV's best-covered sources, and container scans produce a lot of dpkg packages.

## Fixing it

Two independent pieces:

1. **PURL types.** Add explicit cases for `hackage`, `cran`, `opam`, `deb`, and hex-for-erlang. Worth considering whether the fallback should prefer the *most specific* value rather than the first — that alone fixes ocaml. This touches PURL generation globally (SBOM output, dedup keys, cache keys), so it wants its own change and a careful look at what else keys off these strings.
2. **The fallback.** Either delete `ecosystemToOSV` and its branch as dead code, or make it reachable — e.g. try the PURL query first and retry by name+ecosystem when the PURL type isn't one OSV knows. The second is more useful and would make the declared list honest even where the PURL type is off.

## Beyond the current declaration

OSV also covers sources with no Bomly ecosystem to map onto: **Julia**, Bitnami, Android, the Linux kernel, curl, Root, OSS-Fuzz, and git/C++ (commit-range based, distinct from Conan). Julia is the one that looks like a genuine ecosystem worth supporting; the rest are either not dependency ecosystems or out of scope for a manifest scanner. Filing them here for visibility rather than as a request.

Ref: https://google.github.io/osv.dev/data/#covered-ecosystems

bomly-guy:
Fixes #317.

Five ecosystems (`erlang`, `haskell`, `r`, `ocaml`, `dpkg`) were declared as OSV-supported but could never match: Bomly emitted a PURL type outside the purl spec for each, and the name+ecosystem fallback that would have rescued them was unreachable.

## 1. PURL types

`PackageURLTypeForValues` consults an explicit switch for every value and otherwise returns the first non-empty value verbatim, so any ecosystem without an explicit case leaked its Bomly identifier into the PURL. Added explicit cases:

| Ecosystem | Was | Now |
| --- | --- | --- |
| erlang | `pkg:erlang` | `pkg:hex` |
| haskell | `pkg:haskell` | `pkg:hackage` |
| r | `pkg:r` | `pkg:cran` |
| ocaml | `pkg:ocaml` | `pkg:opam` |
| dpkg | `pkg:dpkg` | `pkg:deb` |

Verified against the live OSV API:

```
pkg:cran/commonmark@1.8        -> 1 vuln     pkg:r/commonmark@1.8      -> 0
pkg:hackage/aeson@1.4.0.0      -> 2 vulns    pkg:haskell/...           -> 0
pkg:deb/debian/curl@7.64.0-4   -> 131 vulns  pkg:dpkg/curl@7.64.0-4    -> 0
pkg:hex/coherence@0.5.0        -> 1 vuln
```

The same problem existed one layer down for `rust`, `dotnet`, `dart`, `elixir` and `cpp`: those only mapped correctly because the package manager rescued them, and the manager is not always populated (SBOM ingest, syft-sourced container packages). The ecosystem alone is now enough. `swift` is deliberately left without a case — it is a valid purl type in its own right, and adding one would beat `cocoapods` whenever the ecosystem is checked first.

## 2. SBOM round-trip

SBOM ingest recovers the ecosystem from the purl type via `ParseEcosystem`, which only knows Bomly's own identifiers — so `pkg:deb` would have come back as unknown where `pkg:dpkg` resolved. Added an explicit inverse table in `internal/sbom/identity.go`, which also fixes types that were already broken there (`hex`, `cargo`, `gem`, `pypi`, `nuget`, `composer`, `pub`, `conan`, `cocoapods`).

## 3. The dead fallback

`buildQuery` now uses the name+ecosystem form when the canonical PURL's type is not one OSV indexes (`osvPURLTypes`). Where neither the type nor the ecosystem is known to OSV, the PURL query is still sent — no package loses the query it gets today.

## Known limitation (documented in code)

Distro types need the distro namespace to match: `pkg:deb/debian/curl@...` resolves, `pkg:deb/curl@...` does not. Container and image scans get that namespace from the upstream syft PURL. A bare dpkg package with no PURL of its own still cannot be matched, and there is no name fallback for it either since OSV keys Debian advisories by release (`Debian:12`) rather than by distro alone.

## Not addressed

Julia and OSV's non-dependency sources (Bitnami, Android, Linux kernel, curl, OSS-Fuzz, git/C++) — filed in the issue for visibility rather than as a request.

## Testing

- `TestPackageURLTypeForValuesUsesSpecTypes` — the five fixed mappings plus regression guards for the ones that already worked.
- `TestPackageURLTypeForEcosystemAlone` — every ecosystem with a spec purl type reaches it without a package manager.
- `TestEcosystemFromPURLTypeRoundTripsEmittedPURLs` — every PURL Bomly emits names an ecosystem on the way back in.
- `TestDeclaredEcosystemsAreQueryable` — rewritten: each declared ecosystem must produce either an OSV-indexed PURL type or a name+ecosystem query.
- `TestBuildQueryFallsBackWhenPURLTypeIsNotOSVIndexed` / `TestBuildQueryKeepsPURLWhenNoEcosystemName`.

`make test`, `make lint` (0 issues), and `make generate` (no diff) all clean.


sdk/purl.go
```diff
@@ -82,26 +82,48 @@ func buildPackageURLFallback(purlType, namespace, name, version string) string {
...
 			return "swift"
 		case "github-actions", "githubactions":
 			return "githubactions"
-		case "conan":
+		case "conan", "cpp":
 			return "conan"
-		case "mix", "hex":
+		case "mix", "hex", "elixir", "erlang", "rebar", "otp":
+			// Elixir and Erlang both publish to Hex.
 			return "hex"
```
bomly-guy, line 124: [P2] Do not infer Hex provenance for OTP artifacts

`PackageManagerOTP` is registered for `*.app`, which includes OTP/runtime applications and does not establish that the artifact came from Hex. Because this switch sees `EcosystemErlang` first, every Erlang dependency without an existing PURL—including OTP apps—now becomes `pkg:hex/<name>@<version>`, and OSV will query it as a Hex package. A name/version collision can therefore produce a false vulnerability finding. Please restrict `hex` synthesis to dependencies with known Hex provenance (or at minimum stop mapping `otp`/bare Erlang to Hex).

internal/sbom/identity.go
```diff
@@ -11,11 +11,40 @@ func parsePURL(value string) *packageurl.PackageURL {
...
 
+// purlTypeEcosystems inverts sdk.PackageURLTypeForValues for the purl types
+// whose spec name differs from the Bomly ecosystem name. Without these, a PURL
+// Bomly itself emitted would not round-trip through SBOM ingest: ParseEcosystem
+// only knows Bomly's own identifiers.
+var purlTypeEcosystems = map[string]sdk.Ecosystem{
+	"golang": sdk.EcosystemGo,
+	// Elixir and Erlang share pkg:hex; Elixir is the far more common source of
+	// a hex PURL, and the component's own ecosystem field wins when present.
+	"hex":       sdk.EcosystemElixir,
```
bomly-guy: [P2] Preserve Erlang identity across SBOM round-trips

The standard codecs do not serialize `Component.Ecosystem`: CycloneDX drops it, and SPDX reconstructs it from the PURL. Consequently an Erlang dependency emitted as `pkg:hex/...` is re-imported here as Elixir, and `packageManagerForPURLType` then labels it Mix. Before this change, `pkg:erlang/...` at least round-tripped as Erlang, so the new spec-compliant PURL introduces a metadata regression. Since a Hex PURL cannot distinguish Erlang from Elixir, please preserve an explicit ecosystem hint in the encoded SBOM (or leave ambiguous Hex identity unset) and add an Erlang encode/decode round-trip test; the current test only checks that the result is non-unknown.

bomly-guy (commented):
Found two correctness issues in the new Erlang/Hex registry inference.

sdk/purl.go
bomly-guy: Confirmed and fixed in 2a93d03 — you were right, and Syft agrees: `syft/pkg/cataloger/erlang/package.go` emits `packageurl.TypeHex` for `rebar.lock` entries but `packageurl.TypeOTP` for `*.app`, so the ecosystem-level case was overriding a distinction that already existed upstream.

- `hex` is now mapped at the package-manager level only (`mix`, `rebar`).
- `otp` gets its own explicit case, so it survives the value-order rule instead of falling through to `pkg:erlang`.
- A bare `erlang` value with no manager keeps the non-spec `pkg:erlang` rather than guessing a registry the package may not be published to. It is exempted from `TestPackageURLTypeForEcosystemAlone` alongside `swift`, with the reason recorded.

The same unsafe inference existed on the OSV side, where `ecosystemToOSV` mapped `erlang` to `Hex` — my new fallback would have reintroduced the false-match risk by name even after the PURL was fixed. That case is gone too.

Erlang coverage is unaffected: rebar dependencies carry `pkg:hex` and match on it. Against the live API, `pkg:hex/cowboy@2.10.0` returns 2 advisories and `pkg:otp/kernel@9.2` returns 0.

New test `TestBuildQueryDoesNotQueryOTPApplicationsAsHex` pins both halves.

internal/sbom/identity.go
bomly-guy: Fixed in 2a93d03. I took the "leave ambiguous Hex identity unset" option:

- `hex` is out of `purlTypeEcosystems`. It restores the pre-PR behaviour for that type (`ParseEcosystem("hex")` already failed, so it was unknown before too) and stops the confident mislabel — unknown is the only honest answer when the PURL genuinely cannot distinguish rebar from mix.
- `otp` is mapped to `EcosystemErlang`. Unlike `pkg:hex` it names exactly one ecosystem, so OTP applications now round-trip as Erlang rather than being lost.
- `maven` stays, with the reason recorded inline: it is ambiguous the same way, but `ParseEcosystem` already resolved it before this table existed, so dropping it would regress every Java SBOM to unknown. That is a pre-existing question rather than one this PR introduces.

Added `TestEncodeDecodeRoundTripPreservesErlangIdentity`, which encodes an Erlang dependency and decodes it back through both SPDX 2.3 and CycloneDX 1.6, asserting the OTP app survives as `erlang` and that `pkg:hex` comes back unknown and is never labelled `mix`. I verified it fails on the old behaviour before keeping it:

```
--- FAIL: .../rebar_dependency/cyclonedx-1.6+json
    sbom_test.go:831: ecosystem = "elixir", want ""
    sbom_test.go:834: ambiguous pkg:hex must not be labelled "mix"
```

I did not add a `bomly:` ecosystem hint to the encoded SBOM. The mechanism exists (SPDX `PackageComment` fields, CycloneDX properties), but CycloneDX has no property-decode path today, so wiring it up is a codec change with its own round-trip surface — worth its own PR rather than riding along on the OSV fix. Happy to do it as a follow-up if you would rather have the hint than the unknown.

PR #322, opened 2026-07-25
fix(matchers): query advisories by the ecosystem-native package name

### LINKED ISSUE #319: Scoped npm packages match advisories of the unscoped package (grype matcher drops the scope)
## Summary

Scoped npm packages inherit every advisory of the same-named **unscoped** package. `@tailwindcss/postcss@4.3.3` is reported as vulnerable to postcss advisories; `@scope/lodash` picks up lodash advisories, and so on. The finding is attached to the scoped package's PURL, so the remediation output then recommends a version that does not exist for that package.

## Root cause

For npm, the scope lives in `Coordinates.Org` and `Coordinates.Name` holds the **bare** name (`Org: "tailwindcss"`, `Name: "postcss"`). `graphPkgToGrypePkg` passes `p.Name` straight through:

```go
// internal/matchers/grype/builtin.go:105
func graphPkgToGrypePkg(p *sdk.Package) grypepkg.Package {
	return grypepkg.Package{
		ID:       grypepkg.ID(p.PURL),
		Name:     p.Name,          // <-- "postcss", not "@tailwindcss/postcss"
		...
	}
}
```

Grype searches its DB strictly by the name it is handed (`db/v6/name.PackageNames` → `search.ByPackageName`), with no scope reconstruction from the PURL. So the query for `@tailwindcss/postcss` is really a query for `postcss`, and every postcss advisory matches. The PURL we pass is correct, which is why the bogus results get attached back to the scoped package via `applyMatches`.

`Coordinates.DisplayName()` already produces the ecosystem-native form (`@org/name` for npm), which is what grype and syft expect.

## Reproduction

Minimal project with a scoped package that does not exist upstream:

`package-lock.json`
```json
{
  "name": "scopetest", "version": "1.0.0", "lockfileVersion": 3, "requires": true,
  "packages": {
    "": {"name":"scopetest","version":"1.0.0","dependencies":{"@fake-scope/lodash":"1.0.0"}},
    "node_modules/@fake-scope/lodash": {"version":"1.0.0","resolved":"https://registry.npmjs.org/@fake-scope/lodash/-/lodash-1.0.0.tgz","integrity":"sha512-x"}
  }
}
```

```sh
bomly scan --path . --enrich --matchers grype --format json
```

Actual:

```
pkg:npm/%40fake-scope/lodash@1.0.0  @fake-scope/lodash  5 vulns
  GHSA-35jh-r3h4-6jhm, GHSA-4xc9-xhrj-v574, GHSA-f23m-r3pf-42rh, GHSA-fvqr-27wr-82fm, GHSA-jf85-cpcp-j695
```

Expected: 0 vulnerabilities.

Real-world instance (bomly-landing-page, v0.20.1): `@tailwindcss/postcss@4.3.3` reported with `GHSA-6g55-p6wh-862q` (fixed in 8.5.12) and `GHSA-r28c-9q8g-f849` (fixed in 8.5.18) — postcss advisories and postcss fix versions — while the project's actual `postcss@8.5.16` is reported clean. These two false positives fail `bomly diff --fail-on high` in Bomly Guard on every PR that bumps the tailwind package: https://github.com/bomly-dev/bomly-landing-page/actions/runs/30154178376/job/89669424317

## Impact

- **False positives** on every scoped npm package whose bare name collides with a vulnerable unscoped package — a large surface in the npm ecosystem (`@tailwindcss/postcss`, `@types/*`, `@babel/*`, `@rollup/*`, …).
- **Bogus remediation**: the derived fix suggests bumping `@tailwindcss/postcss` to `8.5.18`, a version that does not exist for that package.
- **False negatives are possible in the other direction** too: a genuinely vulnerable scoped package is queried under the wrong name and its own advisories are never found.
- CI gating (`--fail-on`) and SARIF alerts inherit all of the above.

## Proposed fix

- Pass the ecosystem-native name to grype: use `p.DisplayName()` instead of `p.Name` in `graphPkgToGrypePkg` (`internal/matchers/grype/builtin.go:105`). Confirm the same for external grype mode (`internal/matchers/grype/external.go`).
- Audit the OSV matcher for the same class of bug — the name-keyed lookup path uses the bare name (`internal/matchers/osv/matcher.go:522`, `cache.NewKey("", dep.Name, ...)` / `NamePackage{Name: dep.Name}`). OSV expects `@scope/name` for npm, and the cache key is also under-specified without the scope.
- Add regression tests covering a scoped npm package: (a) it does not match advisories of the same-named unscoped package, (b) it does match advisories keyed under its own scoped name.
- Consider whether any other matcher/analyzer consumes `Package.Name` where the ecosystem-native name is required — a single audit pass over `.Name` uses in `internal/matchers/**` would be worthwhile.

Found while investigating a Bomly Guard failure on bomly-landing-page#245. Verified against `bin/bomly` built from `main` (v0.20.1 line).

bomly-guy:
Fixes #319.

## The bug

`Coordinates` splits identity into `Org` + `Name` following the PURL namespace/name split, so `Name` alone is `postcss` for both `postcss` and `@tailwindcss/postcss`. Grype searches its DB strictly by the name it is handed — it reconstructs a namespace only for Java, from the PURL, in its own resolver — so `graphPkgToGrypePkg` passing `p.Name` made every scoped npm package inherit the same-named unscoped package's advisories, attached to the scoped PURL, with remediation pointing at versions that do not exist for it.

## The fix

New `Coordinates.EcosystemName()` — the identity-grade counterpart to the presentation-only `DisplayName()` — rebuilds the ecosystem-native form: `@org/name` for npm, `org:name` for the Maven family, `org/name` for everything else (Go, Composer, Swift, GitHub Actions). It is now used everywhere a package name leaves the process:

| Site | Was | Impact |
| --- | --- | --- |
| `matchers/grype/builtin.go` | `p.Name` | The reported bug: scoped packages matched unscoped advisories. |
| `sbom/transform.go` | `pkg.QualifiedName()` | Emitted `tailwindcss:postcss`, which is not a name in any ecosystem. External grype mode feeds this SPDX document to the grype CLI, so scoped packages matched **nothing** there. Also a plain SBOM-correctness fix. |
| `matchers/osv/matcher.go` | `dep.Name` | Name-keyed query + cache key (the query path is currently unreachable per #317; the key was under-specified either way). |
| `matchers/depsdev/matcher.go` | `pkg.Name` | Name-fallback cache key only — `depsDevName` already handled scopes, but the key collided across them. |
| `analyzers/jsreach/analyzer.go` | `QualifiedName()`, `Name` | Import seeds. `QualifiedName()` never matched a real specifier, and the bare `Name` seeded `@scope/pkg` as reachable from any `import "pkg"`. |

Maven is unaffected in practice — grype's `JavaResolver` derives `group:artifact` from the PURL and discards `p.Name` — but the value is now correct at the source rather than incidentally.

## Verification

Against the issue's reproduction, with `bin/bomly` built from this branch (`--enrich --matchers grype`, real DB):

```
pkg:npm/%40fake-scope/lodash@1.0.0  @fake-scope/lodash   0 vulns   (was 5)
pkg:npm/lodash@4.17.15              lodash               6 vulns   (unchanged)
```

And the other direction — a genuinely vulnerable scoped package still resolves under its own name:

```
pkg:npm/%40babel/traverse@7.23.1    @babel/traverse      1 vuln    GHSA-67hx-6x53-jw92
```

## Tests

- `sdk`: `EcosystemName` table across npm (scoped, already-`@`-prefixed, unscoped, package-manager-only), Maven, Scala, Go, Composer, Swift, GitHub Actions, and the empty-org/empty-name edges.
- `grype`: name mapping per ecosystem, plus a test that runs grype's own `db/v6/name.PackageNames` over the mapped package to assert the DB is searched for `@tailwindcss/postcss` and never `postcss` — this closes the loop without needing a DB.
- `osv`: scoped query name, and that scoped/unscoped no longer share a cache key.
- `sbom`: component names are ecosystem-native.
- `jsreach`: the existing scoped test now feeds the runner a real bare specifier (its previous `QualifiedName()` fixture was a form the runner never emits); a new test asserts a scoped package is *not* seeded by an unscoped import.

`make test`, `make lint`, and `make generate` are clean, and the build/tests pass under `bomly_external_grype,bomly_external_syft`.

## Notes for review

- No smoke case is added. Covering this needs a real grype DB plus a pinned public repo with a scoped package whose bare name collides with a vulnerable unscoped one; happy to add one if you want to point at (or create) a fixture repo.
- No golden churn expected: no scoped npm package in any current golden carries vulnerabilities, and no smoke case compares a *generated* SBOM.
- `DisplayName()` is left alone. It produces a similar string but is documented as presentation-only, and changing it would move user-visible JSON `name` fields (e.g. `actions:checkout` → `actions/checkout`) — a separate call, not a bug fix. The distinction between the three name accessors is now written down in `dev-docs/MODELS.md`, with a decision-log entry in `dev-docs/ARCHITECTURE.md`.


sdk/coordinates.go
```diff
@@ -49,6 +49,29 @@ func (i Coordinates) DisplayName() string {
...
+	if org == "" || name == "" {
+		return qualifiedName(org, name)
+	}
+	switch i.displayEcosystem() {
+	case EcosystemNPM:
+		return "@" + strings.TrimPrefix(org, "@") + "/" + name
+	case EcosystemMaven, EcosystemScala:
+		return org + ":" + name
+	default:
+		return org + "/" + name
```
bomly-guy, line 80: [P1] Preserve bare names for OS packages

This default also applies to APK/DPKG/RPM/etc., where `Org` comes from the PURL namespace (for example `alpine`) but is not part of the package name Grype queries. For a Syft coordinate with `Org: "alpine"`, `Name: "libcrypto3"`, and `PURL: "pkg:apk/alpine/libcrypto3@..."`, this now makes `graphPkgToGrypePkg` set `Package.Name` to `alpine/libcrypto3`; Grype's APK matcher calls `search.ByPackageName(p.Name)`, so the distro advisory keyed as `libcrypto3` is missed. I confirmed this with a focused test on the PR head (`Name = "alpine/libcrypto3", want "libcrypto3"`). `FromDepGraph` emits the same name into the SPDX consumed by external Grype. Please restrict the join to ecosystems whose native package name actually includes the namespace and preserve the bare `Name` for OS ecosystems; an APK case alongside the #323 distro/upstream regression test would cover both modes.

bomly-guy (commented):
Found one blocking regression in the cross-ecosystem fallback.

sdk/coordinates.go
bomly-guy: Good catch — fixed in 2ad5b5e, this was a real regression.

You're right that `Org` is the distro for OS packages, and the `org/name` default would have traded the npm false positives for missing every OS advisory — precisely the data #323 just plumbed distro and upstreams through to reach.

Joining is now **opt-in per ecosystem** rather than the default. `EcosystemName()` joins only npm (`@org/name`), the Maven family (`org:name`), and Go / Composer / Swift / GitHub Actions (`org/name`) — the ecosystems whose advisory databases actually key on the namespaced form — and returns the bare `Name` for everything else. That covers APK/DPKG/RPM/ALPM/portage/homebrew, and also conan and any other ecosystem whose PURL namespace names a vendor or channel.

Tests, as suggested:
- `sdk`: APK, DPKG, RPM, and conan rows added to the `EcosystemName` table.
- `grype`: APK and DPKG cases in the name-mapping test, plus `TestGraphPkgToGrypePkg_OSPackageKeepsBareNameAndDistro`, which pins the bare name, the distro, and the upstreams together on one `pkg:apk/alpine/libcrypto3` coordinate so the two mechanisms can't regress each other. It also asserts `grypeName.PackageNames` resolves to `[libcrypto3]`, matching the shape you used to confirm the bug.

Verified against a live DB rather than only in unit tests: `scan --image alpine:3.17 --enrich --matchers grype` returns the identical set of 26 (purl, advisory) pairs on this branch and on `main` at 868a5f3, and the issue's scoped-npm reproduction still reports 0.

The rule is now written down in both `dev-docs/MODELS.md` and the `ARCHITECTURE.md` decision entry: adding an ecosystem to the join list is a claim about how its advisory database keys packages, and belongs with a test.

PR #325, opened 2026-07-25
Honor Bomly no-proxy with standard proxy fallback

bomly-guy:
## What changed

- Apply `BOMLY_HTTP_NO_PROXY` when Bomly uses `HTTP_PROXY` or `HTTPS_PROXY` from the standard environment.
- Preserve standard proxy routing for hosts that do not match the Bomly bypass list.
- Add a regression test covering both bypassed and proxied requests.

## Root cause

When no Bomly-specific proxy URL or host was configured, the shared HTTP client returned Go's environment proxy function immediately. That path honored the standard `NO_PROXY` value but never applied `BOMLY_HTTP_NO_PROXY`.

## User impact

Users can combine a standard environment proxy with Bomly's explicit no-proxy list. Matching internal or on-premises hosts now bypass the proxy as configured.

## Validation

- `go test ./sdk`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved outbound HTTP proxy handling by combining configured bypass hosts with standard environment settings.
  * Ensured configured no-proxy hosts correctly bypass both environment-provided and explicitly configured proxies.
  * Preserved standard proxy behavior when no custom proxy or bypass settings are provided.

* **Documentation**
  * Clarified how configured proxy and no-proxy values interact with standard environment settings.

### DIFF
--- a/docs/CONFIG_REFERENCE.md
+++ b/docs/CONFIG_REFERENCE.md
@@ -54,8 +54,8 @@ YAML files use the nested keys documented below. Unknown keys and the former fla
 | `pipeline.install_args` | `BOMLY_INSTALL_ARGS` | `[]string` | - | Additional detector-specific install arguments |
 | `logging.quiet` | `BOMLY_QUIET` | `bool` | - | Suppress all non-error output |
 | `logging.verbosity` | `BOMLY_VERBOSE` | `int` | - | Verbosity level (0=normal, 1=verbose, 2+=debug) |
-| `network.proxy.url` | `BOMLY_HTTP_PROXY` | `string` | - | Outbound HTTP proxy URL used by Bomly and managed plugins |
-| `network.proxy.no_proxy` | `BOMLY_HTTP_NO_PROXY` | `string` | - | Comma-separated hosts, domains, or CIDRs that should bypass the outbound HTTP proxy |
+| `network.proxy.url` | `BOMLY_HTTP_PROXY` | `string` | - | Outbound HTTP proxy URL; when set, it replaces standard HTTP_PROXY/HTTPS_PROXY URLs |
+| `network.proxy.no_proxy` | `BOMLY_HTTP_NO_PROXY` | `string` | - | Hosts, domains, or CIDRs added to the standard NO_PROXY/no_proxy bypass list |
 | `network.proxy.type` | `BOMLY_HTTP_PROXY_TYPE` | `string` | http | Outbound proxy type when using host/port proxy settings: http, https, or socks5 |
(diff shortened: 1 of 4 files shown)


bomly-guy:
Reviewed — the fix is correct and the bug it closes is real: before this, setting `BOMLY_HTTP_NO_PROXY` with no Bomly proxy silently did nothing, and the behavior was inconsistent with the `proxyURL != ""` branch that has always applied `NoProxy`.

One behavioral note worth documenting (not a blocker):

`envProxy.NoProxy = noProxy` **replaces** the environment's `NO_PROXY` rather than adding to it. So a user who has a working `NO_PROXY=.corp.example` and then sets `BOMLY_HTTP_NO_PROXY=.internal.example` loses the `.corp.example` bypass for Bomly's own requests only — a surprising, hard-to-diagnose partial failure. This is consistent with the explicit-proxy branch (which already replaces), so if replace-wins is the intended precedence rule, it should be stated in [`docs/CONFIG_REFERENCE.md`](docs/CONFIG_REFERENCE.md) next to `network.proxy.no_proxy` alongside the existing "a Bomly proxy setting takes precedence over the standard variables" language. Merging the two lists (`envProxy.NoProxy + "," + noProxy`) would be the less surprising behavior, but either way it should be an explicit documented choice rather than an implementation detail.

Minor: this branch switches from `http.ProxyFromEnvironment` (which caches env at first use via `sync.Once`) to a live `httpproxy.FromEnvironment()` read per client construction. That is fine and arguably better, but it does mean the two branches now differ in when environment changes take effect. No action needed.

bomly-guy:
Re-reviewed at `Merge standard and Bomly proxy bypass lists`. Resolved, and more thoroughly than I suggested — `mergeNoProxy` applies to the explicit-proxy branch too, so the two paths now share one precedence rule instead of both silently replacing. `TestMergeNoProxyPreservesOrderAndRemovesDuplicates` pins order-preserving case-insensitive dedup, `TestNewHTTPClientMergesNoProxyWithExplicitProxy` covers the branch I hadn't asked about, and the `doc:` tags in `internal/config/config.go` were updated so the generated [docs/CONFIG_REFERENCE.md](docs/CONFIG_REFERENCE.md) now states the actual semantics for both `network.proxy.url` (replaces) and `network.proxy.no_proxy` (adds). Nothing further from me.

Note the test rename to `TestNewHTTPClientMergesNoProxyWithStandardProxyFallback` — [#337](https://github.com/bomly-dev/bomly-cli/pull/337) cited the old name in its inventory and has already been updated to match, so no stale reference there.

PR #326, opened 2026-07-25
Verify filesystem containment boundaries

bomly-guy:
## What changed

- verify that Git target materialization checks out the requested commit without moving the source repository
- record that repository symlinks remain symlinks during materialization
- exercise ZIP and tar rejection for traversal paths, absolute paths, links, and special files
- prove package identity text is hashed before it becomes a cache filename
- prove generated documents write only to the explicitly selected output destination
- document the filesystem-containment assurance scope and explicit-path trust boundary
- pass the default archive limits through containment tests so they exercise the bounded extraction API

## Why

These tests turn filesystem and archive safety claims into repeatable evidence. They also keep explicit user-selected paths distinct from paths selected automatically from repository content.

This is an assurance-only change. It does not alter runtime behavior.

## Validation

- focused Git, plugin, cache, output, baseline, and option tests
- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`

## Merge order

This PR is stacked on #332 because its containment tests use the bounded archive extraction API. Merge in this order: #336, #332, then #326. After #332 merges, rebase this branch onto `main` and retarget the PR before merging.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards that reject oversized configuration, baseline, SBOM, plugin, and network response inputs with clear limit errors.
  * Improved plugin archive extraction to prevent unsafe paths, unsupported file types, partial files, and excessive expanded content.
  * Prevented upstream error responses from exposing remote response bodies.
  * Improved baseline duplicate detection and validation consistency.
  * Ensured generated output is written only to the explicitly selected destination.

* **Documentation**
  * Documented input, archive, network, caching, and security limits across user and developer guides.


bomly-guy:
Reviewed. Test-only, and the cases chosen are the ones that matter: the zip/tar tables cover parent traversal, absolute paths, Windows-style `..\`, symlink mode, hard links, and character devices — the tar special-file cases in particular are the ones a hand-rolled extractor usually misses. `TestFileCache_HostileIdentityCannotEscapeCacheDirectory` asserting both the derived path stays under the root *and* that nothing was actually written outside is the right belt-and-braces shape. `TestMaterializeLocalRefKeepsRepositorySymlinksAsSymlinks` pins a property that's easy to break by "improving" materialization to copy files.

**Blocking-ish: this conflicts with [#332](https://github.com/bomly-dev/bomly-cli/pull/332).** That PR changes `extractZipArchive` / `extractTarGzArchive` to take a third `archiveLimits` parameter. `internal/plugin/install_containment_test.go` here calls both with two arguments, so whichever lands second won't compile. This is the smaller change, so the cheapest resolution is to merge this first and fix up the two call sites in #332, or land this with `defaultArchiveLimits()` if #332 goes first. Either way it needs deciding now rather than in CI.

**On [test/assurance/SECURITY_CONTAINMENT.md](test/assurance/SECURITY_CONTAINMENT.md):** the doc lists six boundaries but this PR only adds tests for four of them. "Recursive project discovery does not follow symlinked directories" and the baseline claims ("automatic selection uses the conventional path", "malformed documents fail closed", "replacement does not expose partial data") are covered by tests in other packages or other PRs in this batch. That's fine, but the doc as written reads like an index to tests in this PR, so a reader can't tell which claims are proven here versus asserted. [#329](https://github.com/bomly-dev/bomly-cli/pull/329)'s `NETWORK_BOUNDARIES.md` solves exactly this with a `| Boundary | Expected behavior | Main coverage |` table naming the test files — adopting the same format here (and in [#328](https://github.com/bomly-dev/bomly-cli/pull/328)'s `EXECUTION_BOUNDARIES.md`) would make all three consistent and make a stale claim visible when a test is renamed or deleted.

The "Explicit paths are authority boundaries, not project sandboxes" paragraph is the right framing and worth keeping verbatim — it's the distinction that makes the [#330](https://github.com/bomly-dev/bomly-cli/pull/330) baseline-symlink behavior coherent.

bomly-guy:
Re-reviewed at `docs: map filesystem containment boundaries to tests`. The claim→test table is exactly right, and it's better than what I suggested: pulling in `TestPlanSubprojectsRecursiveDoesNotFollowSymlinkedDirs`, `TestLoadRejectsMalformedAndUnsupportedDocuments`, and `TestWriteAtomicValidationFailurePreservesExistingDocument` from other packages means every claim in the doc now names the test that proves it, including the ones this PR doesn't add. [#329](https://github.com/bomly-dev/bomly-cli/pull/329) and [#328](https://github.com/bomly-dev/bomly-cli/pull/328) have converged on the same format.

**The [#332](https://github.com/bomly-dev/bomly-cli/pull/332) conflict is still live and will break the build.** [internal/plugin/install_containment_test.go:49](internal/plugin/install_containment_test.go:49) and [:106](internal/plugin/install_containment_test.go:106) still call:

```go
err = extractZipArchive(archivePath, t.TempDir())
err := extractTarGzArchive(archivePath, t.TempDir())
```

#332 is at `fix: close managed plugin resource limit gaps` and both functions still take the third `archiveLimits` parameter. Whichever of the two merges second fails to compile — this is a compile error in the merge result, not a textual conflict git will flag, so it won't surface until CI runs on the second PR (or on `main`, if both are queued).

Cheapest resolution: add `defaultArchiveLimits()` to both call sites here and merge this one after #332. That works whether or not #332 moves, since the parameter already exists on its branch.

bomly-guy:
Re-reviewed at `test(plugin): use bounded archive extraction`. Resolved, and better than the fix I suggested: rather than just adding `defaultArchiveLimits()` and relying on someone remembering the merge order, the PR is now stacked — base branch retargeted to `plugin-download-limits`, so [#332](https://github.com/bomly-dev/bomly-cli/pull/332)'s `install.go` is in the history and the arity mismatch is structurally impossible instead of merely avoided. GitHub enforces the ordering now.

Verified on this head: `go build ./...` succeeds and `go test ./internal/plugin ./internal/baseline ./internal/system ./internal/matchers/... ./internal/cli/render` all pass across the full #336 → #332 → #326 stack. CI is green.

Nothing outstanding from me.

PR #327, opened 2026-07-25
Redact credentials from endpoint logs

bomly-guy:
## What changed

- Remove URL user information before writing the configured Scorecard API endpoint to debug logs.
- Replace malformed endpoint text with a fixed marker rather than logging the original value.
- Add a registry-level regression test that captures the debug event and proves credentials are absent.

## Root cause

The Scorecard matcher registration logged the configured API base verbatim. An explicitly trusted endpoint containing URL credentials therefore exposed its username and password at `-vv`.

## User impact

Debug logs retain the endpoint scheme, host, path, cache directory, and cache lifetime needed for troubleshooting without exposing URL credentials.

## Validation

- `go test ./internal/registry`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`

### DIFF
--- a/internal/registry/builder.go
+++ b/internal/registry/builder.go
@@ -3,6 +3,7 @@ package registry
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"sort"
 	"strconv"
 	"strings"
@@ -389,10 +390,19 @@ func (r *Registry) registerScorecardMatcher() {
 		r.RegisterMatcherWithOptions(matcher, ComponentOptions{DefaultEnabled: false})
 	}
 	r.logger.Debug("scorecard matcher configured",
-		zap.String("api_base", scoreCfg.APIBase),
+		zap.String("api_base", endpointForLog(scoreCfg.APIBase)),
 		zap.String("cache_dir", scoreCfg.CacheDir),
 		zap.Duration("cache_ttl", scoreCfg.CacheTTL),
 	)
+}
+
+func endpointForLog(value string) string {
+	parsed, err := url.Parse(value)
+	if err != nil {
+		return "[invalid URL]"
+	}
+	parsed.User = nil
+	return parsed.String()
 }
 
 func (r *Registry) httpClientProvider() *sdk.HTTPClientProvider {
--- a/internal/registry/builder_test.go
+++ b/internal/registry/builder_test.go
@@ -2,11 +2,14 @@ package registry
 
 import (
 	"slices"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/bomly-dev/bomly-cli/sdk"
 	"go.uber.org/zap"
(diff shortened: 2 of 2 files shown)


bomly-guy:
Reviewed — correct fix, good test (asserting the exact redacted string rather than just "doesn't contain the secret" is the right shape).

Two notes:

**1. `endpointForLog` duplicates a helper [#334](https://github.com/bomly-dev/bomly-cli/pull/334) is adding.** That PR introduces `logging.SanitizeURL` / `redactURLUserinfo`, which does the same job. Two independent URL-redaction helpers in the tree is exactly how one of them drifts out of date. Suggest one of: land this as-is and follow up by replacing `endpointForLog` with `logging.SanitizeURL`, or rebase on #334 and use it directly. Note the two differ in behavior — yours drops userinfo entirely and returns `"[invalid URL]"` on parse failure; `SanitizeURL` rewrites userinfo to `[REDACTED]` and returns the input unchanged on parse failure. Returning the raw input on a parse error is the weaker behavior (a malformed URL with a token in it still gets logged), so `endpointForLog`'s fail-closed choice is the better one and worth preserving in whichever helper survives.

**2. The scorecard `api_base` was not the only endpoint-shaped value logged.** `logResolvedOptions` in [internal/cli/root_cmd.go:204](internal/cli/root_cmd.go:204) logs `resolved.URL` verbatim, and a `--url` remote Git target legitimately carries `https://user:token@host/repo`. `internal/git/git.go:148` logs the clone `source` the same way. Both are the same class of leak as this fix. [#334](https://github.com/bomly-dev/bomly-cli/pull/334) covers both, so no change needed here — just flagging so this PR isn't mistaken for complete endpoint-log coverage on its own. Worth checking the merge order so there is no window on `main` where the assurance doc in [#337](https://github.com/bomly-dev/bomly-cli/pull/337) claims coverage that hasn't landed.

bomly-guy:
Re-reviewed. No new commits here, but both of my notes were resolved on [#334](https://github.com/bomly-dev/bomly-cli/pull/334) instead, which is the better place for them:

- `endpointForLog` is gone from `internal/registry/builder.go`, replaced by `logging.SanitizeURL` — and #334 changed `SanitizeURL` to fail closed on a parse error, so the safer behavior from this PR's helper is the one that survived rather than being lost in the consolidation.
- `resolved.URL` in `logResolvedOptions` and the git clone `source` are both sanitized there, with `TestCloneIntoRedactsUserinfoAndPreservesExitError` covering the clone path.

So this PR's `endpointForLog` will disappear in the merge with #334. Worth confirming the sequencing (or rebasing this onto #334) so there's no window on `main` where `builder.go` references a helper that the other branch deleted. Otherwise nothing outstanding.

PR #328, opened 2026-07-25
Add execution boundary assurance

bomly-guy:
## What changed

- verify managed plugins receive only selected protocol, plugin, and proxy settings instead of unrelated host secrets
- prove disabled native plugins do not join the runtime registry
- exercise scan, explain, and diff MCP calls with hostile text while enrichment, audit, and analysis remain off by default
- add a static guard that keeps canonical remediation and built-in hint code free of network, process, cache, and filesystem imports
- document the tested execution and trust boundaries in plain language

## Why

Issue 026 needs repeatable evidence for behavior that crosses process and protocol boundaries. These checks make the current plugin, MCP, and read-only remediation guarantees explicit without changing production behavior.

The inventory also found two production concerns that are intentionally excluded from this assurance-only PR: MCP adapter errors are returned verbatim to clients, and subprocess argv logging is not consistently complete or secret-safe. They will be addressed in separate focused fixes before this assurance layer is considered complete.

## Validation

- `go test ./internal/plugin ./internal/mcp ./internal/remediation ./internal/registry ./test/assurance`
- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`

### DIFF
--- a/internal/mcp/mcp_test.go
+++ b/internal/mcp/mcp_test.go
@@ -183,6 +183,63 @@ func TestScanTool_PropagatesScope(t *testing.T) {
 	}
 }
 
+func TestToolsDoNotEnableNetworkOrAnalysisByDefault(t *testing.T) {
+	t.Run("scan", func(t *testing.T) {
+		adapter := &mockAdapter{
+			scanResult: mcp.ScanRunResult{
+				Response: output.ScanResponse{Command: "scan"},
+			},
+		}
+		c := newTestClient(t, adapter)
+		result := callTool(t, c, "bomly_scan", map[string]any{
+			"path": "../../untrusted\npath",
+		})
+		if result.IsError {
+			t.Fatalf("unexpected tool error: %v", result.Content)
+		}
+		if adapter.scanReq.Enrich || adapter.scanReq.Audit || adapter.scanReq.Analyze {
+			t.Fatalf("scan enabled optional work without permission: %#v", adapter.scanReq)
+		}
+	})
+
+	t.Run("explain", func(t *testing.T) {
+		adapter := &mockAdapter{
+			explainResult: mcp.ExplainRunResult{
+				Response: output.ExplainResponse{Command: "explain"},
+			},
+		}
+		c := newTestClient(t, adapter)
+		result := callTool(t, c, "bomly_explain", map[string]any{
+			"package": "pkg:npm/example@1.0.0\nuntrusted",
+		})
+		if result.IsError {
+			t.Fatalf("unexpected tool error: %v", result.Content)
+		}
+		if adapter.explainReq.Enrich || adapter.explainReq.Audit || adapter.explainReq.Analyze {
+			t.Fatalf("explain enabled optional work without permission: %#v", adapter.explainReq)
+		}
+	})
+
+	t.Run("diff", func(t *testing.T) {
+		adapter := &mockAdapter{
+			diffResult: mcp.DiffRunResult{
+				Response: output.DiffResponse{Command: "diff"},
+			},
+		}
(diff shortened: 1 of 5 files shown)


bomly-guy:
Reviewed. Three worthwhile additions here: `TestPluginEnvDoesNotForwardUnrelatedHostEnvironment` pins an allowlist rather than a denylist (asserting `AWS_SECRET_ACCESS_KEY` / `GITHUB_TOKEN` / `DATABASE_URL` are absent *and* that the required protocol values are present, so a future "just forward everything" regression fails loudly); the disabled-plugin check in `TestInstallDevBinaryVerifyEnableDisableAndUninstall` closes the gap between "recorded disabled" and "actually absent from runtime planning", which are different properties; and the MCP default-authority test feeding control characters and traversal text into `path`/`package`/`base` covers the "doesn't panic and doesn't silently enable" pair together.

**The import-boundary test is weaker than it appears.** `TestReadOnlyRemediationSourcesDoNotImportExecutionOrIOPackages` uses `parser.ParseFile(..., parser.ImportsOnly)` on two named files plus any `remediation.go` under `internal/detectors`. That catches only *direct, same-file* imports. `internal/remediation/derive.go` can import a sibling package in `internal/remediation` that imports `os/exec`, and the test passes. It also can't see the rest of each package: another file in `internal/remediation` importing `net/http` is invisible, even though it's the same package and can be called from `derive.go` freely.

Since the claim in [EXECUTION_BOUNDARIES.md](test/assurance/EXECUTION_BOUNDARIES.md) is about the package ("the central derivation code ... do not import network, process, cache, or filesystem packages"), the check should be at package granularity and transitive. `go list -deps ./internal/remediation` and asserting the forbidden set doesn't appear in the closure would actually prove the stated property, and is about the same amount of code. If per-file is intentional (e.g. you want `os` usable elsewhere in the package but not in `derive.go`), then the doc should say "these files" rather than "the central derivation code", and the test should enumerate all files in the package rather than two by name.

Also: `os` is in the forbidden set but the test file itself uses `os.DirEntry` — fine, but note that forbidding `os` outright is a fairly blunt proxy for "filesystem access" and will fire on something innocuous like `os.Getenv` for a feature flag. Worth a comment recording that the strictness is deliberate so the next person doesn't quietly delete the entry.

**Documentation ordering.** "Assurance for this rule is completed only when every command runner uses the shared redaction boundary" makes the subprocess-logging section explicitly conditional on [#334](https://github.com/bomly-dev/bomly-cli/pull/334). Good that it's stated rather than implied — please just make sure #334 lands before or with this, and before [#337](https://github.com/bomly-dev/bomly-cli/pull/337), whose table asserts the completed state.

**Formatting nit:** the `forbidden` map mixes short stdlib keys with long module paths, so gofmt aligns the whole block to the longest key and the stdlib entries read oddly. Splitting into two maps (stdlib / internal) or one `[]struct{path, reason string}` would read better. Cosmetic only.

Finally, the "MCP is not a sandbox" and "an enabled external plugin is still native code" paragraphs are the most valuable text in the PR — they're the claims most likely to be over-read by someone evaluating Bomly for a locked-down environment, and stating the limit plainly is the right call.

bomly-guy:
Re-reviewed at `Strengthen execution boundary assurance`. The import-boundary test now shells out to `go list -deps -json` and checks the full transitive closure of the target packages, which is what the doc's package-level claim actually needs — an indirect `os/exec` reached through a sibling package is now caught, and the "only two named files" blind spot is gone. Failing loudly when `go list` omits the target package is a good touch; a silently-empty dependency set would have made the test vacuous.

Nothing further from me on this one.

PR #329, opened 2026-07-25
Verify HTTP trust boundaries

bomly-guy:
## Summary

- add live assurance tests for explicit proxy routing and no-proxy host, domain, IP, and CIDR matching
- verify additional CA trust, private-network redirects, cross-host credential stripping, and password-safe transport errors
- prove that private endpoints, proxy settings, CA files, and plugin endpoints require a trusted explicitly selected project config
- document the supported custom-endpoint and private-network boundary in plain language

This PR is test and documentation only. Production defects found during the inventory are intentionally being handled in separate focused fixes.

## User impact

The supported behavior is now explicit: users may select self-hosted or private OSV and Scorecard services, enterprise proxies, and additional CAs. Bomly does not block private destinations or normal redirects, so these settings must come from a config source or environment that the user trusts. Repository config remains opt-in.

## Validation

- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- focused HTTP, config, registry, plugin, and matcher tests

### DIFF
--- a/dev-docs/ARCHITECTURE.md
+++ b/dev-docs/ARCHITECTURE.md
@@ -443,6 +443,22 @@ Permitted enrichment-time services:
 
 Installed external matcher plugins may use their own documented services.
 
+**Custom network settings are trusted authority.** The OSV and Scorecard base
+URLs may point to public, private, loopback, or plain HTTP services. Proxy
+destinations and additional CA files have the same reach. Bomly supports these
+choices for self-hosted services and enterprise networks; it does not apply a
+private-network block. Only the user config is loaded automatically. A
+repository config must be selected with `--config` or `BOMLY_CONFIG`, and
+network-specific environment variables are also explicit inputs.
+
+The shared SDK HTTP client follows Go's normal redirect rules. Redirects are
+allowed because custom services commonly use them, but sensitive headers are
+not forwarded to a different hostname. Proxy and endpoint passwords must not
+appear in errors or logs. Configured PEM certificates extend the system trust
+roots for the current process rather than replacing them. The executable
+assurance matrix is recorded in
+[`test/assurance/NETWORK_BOUNDARIES.md`](../test/assurance/NETWORK_BOUNDARIES.md).
+
 **Native plugins are trusted processes, not sandboxes.** Installation verifies
 the managed artifact and records it disabled by default. Enabling a plugin
 authorizes a native process with the same user-level filesystem, network,
--- a/dev-docs/ARCHITECTURE.md
+++ b/dev-docs/ARCHITECTURE.md
@@ -453,10 +453,12 @@ network-specific environment variables are also explicit inputs.
 
 The shared SDK HTTP client follows Go's normal redirect rules. Redirects are
 allowed because custom services commonly use them, but sensitive headers are
-not forwarded to a different hostname. Proxy and endpoint passwords must not
-appear in errors or logs. Configured PEM certificates extend the system trust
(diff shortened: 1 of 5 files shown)


bomly-guy:
Reviewed. Good slice. The no-proxy matrix covering exact host / domain suffix / CIDR / bare IP is the right set (CIDR and bare-IP are the two people assume work and often don't), the additional-CA test using `httptest.NewTLSServer`'s own cert through a PEM file exercises the real load path rather than a mock, and `TestHTTPClientTransportErrorDoesNotExposeEndpointPassword` covers a leak channel that's easy to forget — `*url.Error` stringifies the URL, so userinfo in an endpoint would otherwise land in whatever wraps that error. The `NoProxy: "*"` + unreachable-proxy trick to force a direct connection while still constructing the client through the real config path is neat.

Three gaps in the coverage relative to what the docs now claim:

**1. The redirect test only proves the cross-host case.** `TestHTTPClientFollowsRedirectToPrivateDestinationWithoutForwardingCredentials` redirects `127.0.0.1` → `localhost`, so Go strips the header and the assertion passes. But the doc claims "sensitive headers are not forwarded to a *different hostname*", which implies the same-host case forwards them — that half is untested, and it's the half that would silently change if someone later installed a custom `CheckRedirect`. A second case asserting credentials *are* preserved on a same-host redirect would pin the actual contract in both directions.

**2. Scheme downgrade on redirect isn't covered or documented.** Go happily follows `https://` → `http://`, so a compromised or misconfigured custom OSV/Scorecard endpoint can downgrade an enrichment request to plaintext, and nothing here notices. Given the "Intentional limits" section already documents not blocking private destinations and not pinning DNS, an https→http redirect downgrade belongs in that same list — it's the same category of deliberate permissiveness, and right now a reader would reasonably assume it's prevented.

**3. The coverage table cites files, not tests.** Every path I spot-checked exists (`internal/plugin/http_test.go`, `internal/registry/builder_test.go`, `internal/config/validate_test.go`), so nothing is wrong — but "matcher tests under `internal/matchers`" for the custom-endpoint row is vague enough that nobody will notice when it stops being true. Naming the specific test functions, as the [#337](https://github.com/bomly-dev/bomly-cli/pull/337) inventory does, makes staleness detectable. This table's format is otherwise the best of the three assurance docs in this batch — I'd suggest [#326](https://github.com/bomly-dev/bomly-cli/pull/326) and [#328](https://github.com/bomly-dev/bomly-cli/pull/328) adopt it.

Minor: `TestCommandContextInitialize_ExplicitConfigCanSelectPrivateNetworkSettings` also happens to pin that `ca_cert_file` resolves relative to the config file's directory rather than the CWD, which is a genuinely load-bearing behavior that I don't think is asserted anywhere else. Worth calling out in the test name or a comment so it doesn't get dropped as incidental.

bomly-guy:
Re-reviewed at `Clarify trusted redirect behavior`. All three points closed:

- `TestHTTPClientPreservesCredentialsOnSameHostRedirect` pins the other half of the contract, so a future custom `CheckRedirect` can't quietly change either direction.
- HTTPS→HTTP redirect downgrade is now stated in "Intentional limits" with the consequence spelled out ("an HTTP redirect sends later requests without transport encryption"), and mirrored in the [#337](https://github.com/bomly-dev/bomly-cli/pull/337) residual column.
- The custom-endpoint row names `TestMatcherMatchEnrichesRegistry` / `TestMatch_AttachesScorecardToPackages` instead of gesturing at a directory, so staleness is now detectable.

Nothing further from me.

PR #330, opened 2026-07-25
Contain automatic baseline discovery

bomly-guy:
## What changed

- warn and ignore a symbolic-link `.bomly` directory or `baseline.json` file during automatic baseline discovery
- apply the same behavior to local filesystem and materialized Git targets
- preserve explicitly selected baseline paths as trusted user input, including explicit symlinks and files outside the project
- document that inspection rejects discovered links but cannot prevent concurrent path replacement
- add regression coverage for file links, directory links, user-selected project-root links, warnings, and both target kinds

## Why

A repository-controlled symbolic link can redirect an automatically discovered baseline read outside the scanned project. Bomly therefore warns and behaves as though no automatic baseline exists instead of failing the whole scan. Explicit baseline paths are different because the user deliberately selects and trusts them.

## Validation

- focused baseline, options, and CLI tests
- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`

### DIFF
--- a/dev-docs/ARCHITECTURE.md
+++ b/dev-docs/ARCHITECTURE.md
@@ -220,13 +220,17 @@ projects. Discovery happens during normal target preparation: scan and explain
 read the materialized project tree, including repositories cloned through
 `--url`, while Git diff independently reads the base and head trees. A detected
 baseline is logged with its path, entry count, selection mode, and target kind;
-each evaluation logs findings evaluated and accepted. Output receives ordinary
-findings whose policy status may be `suppressed` through
-`Finding.PolicyStatus` / `policy_status`, and no baseline-specific output model
-or pipeline stage exists. Renaming the earlier finding field is an intentional
-breaking output-contract change while the CLI output schema identifier remains
-`1.0` and the compact MCP schema remains `mcp/1`. Protocol-v1 decoding still
-accepts the earlier wire field from existing external auditor plugins.
+automatic discovery rejects a symbolic-link `.bomly` directory or baseline
+file so repository content cannot redirect the read outside the materialized
+target. Explicit baseline paths remain trusted user-selected inputs and may
+refer outside the project or through a symbolic link. Each evaluation logs
+findings evaluated and accepted. Output receives ordinary findings whose policy
+status may be `suppressed` through `Finding.PolicyStatus` / `policy_status`, and
+no baseline-specific output model or pipeline stage exists. Renaming the
+earlier finding field is an intentional breaking output-contract change while
+the CLI output schema identifier remains `1.0` and the compact MCP schema
+remains `mcp/1`. Protocol-v1 decoding still accepts the earlier wire field from
+existing external auditor plugins.
 
 ### Decision: registry matching eligibility is an occurrence-level engine boundary
 
--- a/dev-docs/ARCHITECTURE.md
+++ b/dev-docs/ARCHITECTURE.md
@@ -220,10 +220,12 @@ projects. Discovery happens during normal target preparation: scan and explain
 read the materialized project tree, including repositories cloned through
 `--url`, while Git diff independently reads the base and head trees. A detected
 baseline is logged with its path, entry count, selection mode, and target kind;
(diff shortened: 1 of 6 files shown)


bomly-guy:
Reviewed. The threat model is right — repository content should not be able to redirect an automatic read outside the materialized target — and separating "automatic = untrusted convention" from "explicit `--baseline` = trusted user path" is the correct distinction. The `baselineRoot` extraction is a good drive-by dedup. Tests cover both target kinds, both file and directory links, and the symlinked-project-root allow case, which is the case most likely to be broken by a naive implementation.

**Main question: should a symlink at the conventional path be fatal?**

`ResolversForTarget` errors propagate through [internal/cli/opts/options.go:345](internal/cli/opts/options.go:345) as `exit.InvalidInputError`, so `bomly scan --audit` on a repository whose `.bomly` is a symlink now exits 2 with no scan output at all. Two consequences:

- Repository content it does not control now decides whether an audited scan runs. That is a weaker outcome than the goal of the PR: instead of "repo content cannot redirect the read", we get "repo content can abort the run". For an untrusted `--url` target in CI, that's a denial vector that ignoring the file would not create.
- Symlinking a shared `.bomly` directory is a legitimate monorepo pattern (one policy/baseline dir at the repo root, per-package dirs linked to it). Those repos will start failing with exit 2 and will have to add `--baseline none` — a silent breaking change for a valid layout, and the error message points at `--baseline none` (disable) or `--baseline <path>` (trust) but not at "this used to work".

I know this is consistent with the existing documented choice that a malformed automatically-discovered baseline fails closed for audited commands, and that's a fair argument for keeping it. The distinction I'd draw is that a malformed baseline is an actual defect the user must fix, whereas a symlinked baseline is a valid file that Bomly is declining to trust — and for that case "warn and treat as absent" both closes the redirect and keeps the run useful. If you keep the hard failure, please add it to the breaking-change notes and to [docs/BASELINES.md](docs/BASELINES.md) as an explicit migration note for the shared-directory layout, not just as a safety statement.

**Smaller notes**

1. **TOCTOU.** `validateAutomaticPath` walks with `os.Lstat` and then `Load(path)` opens the file separately. A local attacker who can write into the project between the two can still swap in a symlink. Low severity for the stated model (the repo content is already on local disk, and the read is a JSON parse whose result is bounded to policy-status decisions), but it means the guarantee is "rejects committed symlinks", not "cannot follow a symlink". Worth wording that way in [docs/BASELINES.md](docs/BASELINES.md) rather than the stronger "Bomly rejects links so repository content cannot redirect the automatic read outside the project".

2. `validateAutomaticPath` returns `nil` on the first `os.ErrNotExist` component. Correct — a dangling symlink still `Lstat`s successfully so it is caught before that — but a one-line comment saying so would save the next reader the trip through the symlink semantics.

3. The escape check (`relative == ".."` / prefix / `IsAbs`) is unreachable for the automatic selection, since `ResolvePath` builds the path as `filepath.Join(root, DefaultRelativePath)` with a constant relative path. Harmless defense in depth; fine to keep, but a comment noting it guards against a future change to `DefaultRelativePath` would explain why it exists.

bomly-guy:
Re-reviewed at `fix: ignore linked automatic baselines`. Fully addressed:

- A symlinked automatic baseline now warns and behaves as though no baseline exists, via a dedicated `errAutomaticBaselineSymlink` sentinel, so genuine errors from `validateAutomaticPath` still fail closed while the untrusted-content case degrades. That's the right split — repo content can no longer redirect the read *or* abort the run, and the shared-`.bomly`-directory monorepo layout keeps working.
- Both comments landed: the escape check is marked as defense-in-depth against a future `DefaultRelativePath` change, and the `ErrNotExist` early return now explains that a dangling link is caught by `Lstat` first.
- [docs/BASELINES.md](docs/BASELINES.md) and [docs/AUDITORS.md](docs/AUDITORS.md) describe the warn-and-ignore behavior, and the TOCTOU limit is stated plainly ("it does not protect against another process replacing a path between inspection and reading") instead of the stronger claim.

Nothing further from me.

PR #331, opened 2026-07-25
Redact MCP tool errors

bomly-guy:
## What changed

- replace raw scan, explain, diff, and plugin adapter errors with stable client-facing categories
- distinguish request validation, target preparation, target resolution, pipeline, and plugin inventory failures
- preserve wrapped causes for internal callers without sending their text over MCP
- log only the tool, stable category, and Go error type at the MCP boundary
- return a stable message if compact response encoding fails
- document how to use verbose server logs for safe troubleshooting

## Why

MCP tool handlers previously returned `err.Error()` directly. Errors from paths, subprocesses, URLs, plugin configuration, or upstream services can contain credentials or other local details. MCP results enter an agent's context, so arbitrary internal error text must not cross this boundary.

## User impact

Agents receive short, useful messages such as `scan pipeline failed` or `diff target resolution failed`. Users can run `bomly -vv mcp serve` and reproduce the request to inspect safe stage logs on stderr.

## Validation

- `go test ./internal/mcp ./internal/cli`
- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`


bomly-guy:
Reviewed. The category enum plus `WrapToolError` preserving the cause via `Unwrap` is the right structure, the default case failing safe to `"<tool> failed"` means an unwrapped adapter error can't leak by omission, and `TestWrapToolErrorPreservesInternalCause` guards the part that's easy to regress. `jsonResult` no longer echoing `json.Marshal`'s error text is a nice catch — that one could carry struct field paths.

Three things I'd change before merge:

**1. `error_type` is always `*mcp.toolError`, so the log field is dead.**

```go
zap.String("error_type", fmt.Sprintf("%T", err)),
```

By the time `toolErrorResult` runs, `err` is the wrapper for every categorized path, so this always renders `*mcp.toolError` and tells an operator nothing. `TestToolErrorLogsDoNotExposeAdapterDetails` passes precisely because the field is uninformative. Log the type of the unwrapped cause instead — Go type names are authored by us and carry no user data, and `*exec.ExitError` vs `*fs.PathError` vs `*url.Error` is genuinely the most useful single bit for triage:

```go
zap.String("error_type", fmt.Sprintf("%T", errors.Unwrap(err))),
```

**2. The raw cause is now discarded, which contradicts what this PR documents.**

[docs/MCP.md](docs/MCP.md) tells the user to "run the server with `bomly -vv mcp serve` and reproduce the request to inspect the safe stage logs", and the [#337](https://github.com/bomly-dev/bomly-cli/pull/337) security guide says "detailed internal errors stay in local logs". Neither is true for the errors this PR touches: `toolErrorResult` logs only tool + category + type, and the cause is dropped. For a failure that originates in the adapter rather than in a stage that logged on its own — `config.Validate` rejecting a flag combination, `sdk.ParseScope` rejecting a value, `Prepare` failing to resolve a target — there is now no verbosity at which the reason is recoverable.

The redaction target here is the MCP protocol channel, not the operator's own stderr, and the docs already commit to stderr as the diagnostic surface. Adding one Debug line with `zap.Error(err)` alongside the Warn would keep the client-facing contract intact and make the documented troubleshooting step actually work.

**3. `ToolErrorRequest` collapses validation messages an agent could act on.**

`"scan request is invalid"` gives a calling agent nothing to retry with. The underlying text is mostly our own authored validation prose — the repo convention is explicitly actionable messages like `"--audit requires --enrich"` ([CLAUDE.md](CLAUDE.md) feature checklist) — and an agent that can read that will fix its own call, whereas one that can't will retry the same invalid request. Contrast `invalidRequestResult`, which correctly does keep the field name for missing arguments; the validate path is strictly less useful than the sibling it sits next to.

The safety concern is real, since `Prepare`/validate errors can echo a user-supplied path or URL. But that argues for passing through a known-safe message rather than for dropping the category's detail wholesale: a `RequestError` type carrying an authored, data-free message from `config.Validate`, or a small allowlist of validation reasons, would preserve self-correction without echoing inputs. If you'd rather not take that on in this PR, please at least note it in [docs/MCP.md](docs/MCP.md) so it's a recorded trade-off rather than an accident.

bomly-guy:
Re-reviewed at `Clarify secret-safe MCP diagnostics`. Point 1 fixed — `cause_type` now logs `%T` of the unwrapped cause, and the test asserts `*errors.errorString` rather than merely "doesn't contain the secret", so the field can't silently go back to being vacuous.

Points 2 and 3 you've resolved by documentation rather than code, and both docs are now accurate: [docs/MCP.md](docs/MCP.md) states that stage logs exist only where a component emits them independently and that the MCP error log never holds cause text, and the validation trade-off is written down in both `docs/MCP.md` and the architecture decision. That's what I asked for as the fallback, so I'm not going to press either further — they're recorded decisions now, not accidents.

The residual worth keeping in mind: an adapter-origin failure (`config.Validate`, `sdk.ParseScope`, `Prepare`) still has no diagnostic at any verbosity, so `"scan target preparation failed"` under `-vv` is a dead end for whoever operates the server. If that turns into real support load, a `BOMLY_MCP_VERBOSE_ERRORS`-style local opt-in would close it without touching the protocol contract. Not a blocker.

PR #333, opened 2026-07-25
feat(discovery): explain why each probed manifest candidate was skipped

### LINKED ISSUE #248: Per-detector discovery diagnostics: report why a candidate manifest was skipped
Follow-up from #245 (deferred out of #246).

#246 added a depth-limited discovery probe to "no subprojects discovered" errors (`opts.DescribeDiscovery`): it names the manifest evidence that exists and which package manager it belongs to. What it cannot say is *why* a specific candidate was skipped — that requires instrumenting detector applicability checks (readiness probes, evidence-pattern mismatches, filter exclusions) to record structured skip reasons per candidate.

Goal: `found package.json at ./web (npm) — skipped: excluded by --ecosystems go` / `skipped: npm-native detector not ready (npm not on PATH), no committed lockfile for fallback`.

bomly-guy:
Closes #248.

Follow-up to #246: the "no subprojects discovered" probe already named the manifest evidence that exists and which package manager owns it. It now also says why discovery passed each candidate over.

```txt
Nothing to evaluate: no subprojects discovered ...;
  discovery probe: found package.json at web (npm) — skipped: not scanned without --recursive
  discovery probe: found package.json at . (npm) — skipped: excluded by --ecosystems go
  discovery probe: found package.json at . (npm) — skipped: detector filter excludes every npm detector (npm-detector, npm-native-detector, syft-detector)
```

## Reasons

| Reason | Source check |
| --- | --- |
| `not scanned without --recursive` | non-recursive run, nested candidate |
| `below --max-depth N` | recursive depth cap |
| `excluded by --exclude <pattern>` | exclude glob, matched against the candidate **and every ancestor** (the walk prunes subtrees) |
| `excluded by --ecosystems <list>` / `-<eco>` | `EcosystemFilter` include/exclude |
| `no detector registered for <pm>` | no detector names for the manager in the **unfiltered** registry |
| `detector filter excludes every <pm> detector (...)` | `--detectors` / `--exclude-detectors` emptied the chain |
| `no resolvable <pm> manifest in this directory` | planning rejected the candidate for a non-filter reason (e.g. SBOM path resolution) |
| `<detector> not ready (...); <fallback> not ready (...)` | every link in the planned chain failed its readiness probe |

## Implementation

- New `internal/cli/opts/planning_diagnostics.go`: `discoveryDiagnostics` replays the real planning checks **in planning order** rather than threading a reason channel through discovery, so the hot path is unchanged and allocation-free. The replay runs only on the failure path, over at most `discoveryProbeMaxLines` (8) candidates.
- Registration lookups use the request's unfiltered registry, filter checks use the filtered planning registry, so "no detector registered" and "detector filter excludes …" stay distinguishable.
- Readiness is the one check planning never performs (chains are planned statically, probed at resolve time), so the diagnostics call `Ready` directly under a single 5s context, only for candidates that cleared every filter. A chain with at least one ready link yields no reason.
- `DescribeDiscovery` (no request context to replay) keeps emitting evidence lines with no skip clauses.
- `noSubprojectsError` now takes the planning registry; ecosystem-hint rendering deduplicated into `sortedEcosystemNames`.

## Tests / docs

- `internal/cli/opts/planning_diagnostics_test.go`: one test per reason end-to-end through `PlanSubprojects`, plus readiness unit tests (all links unusable → joined reasons; any link ready → silence) and a `DescribeDiscovery` no-speculation test.
- `make test` and `make lint` clean.
- Docs: new "When nothing is discovered" section in `docs/SCAN_TARGETS.md` (reason → fix table), updated MCP troubleshooting entry, and a decision-log entry in `dev-docs/ARCHITECTURE.md` noting the replay trade-off (a reason is re-derived, not recorded — moving a check means moving its replay).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved “No subprojects discovered” errors with a clear summary of the search scope, active filters, manifest candidates, and why each candidate was skipped.
  * Added actionable guidance for recursion, depth limits, exclusions, ecosystem filters, and detector filters.
  * Improved detector readiness errors by identifying unavailable detectors and reporting their readiness reasons.
  * Preserved distinct reporting for toolchain resolution failures.

* **Documentation**
  * Added troubleshooting and scan behavior guidance for discovery and detector errors.


internal/cli/opts/planning_diagnostics.go
```diff
@@ -0,0 +1,223 @@
...
+		return reason
+	}
+	if reason := d.ecosystemSkipReason(manager); reason != "" {
+		return reason
+	}
+	subproject, reason := d.detectorSkipReason(dir, manager)
+	if reason != "" {
+		return reason
+	}
+	return d.readinessSkipReason(subproject, manager)
```
bomly-guy: [P2] Move readiness reporting to the resolution-failure path. `skipReason` runs only from `noSubprojectsError`, after planning returned zero; reaching this line means the same `plannedSubprojectForPackageManager` replay just succeeded with the same scope and filters, in which case the original planner would have returned that subproject and never emitted `noSubprojectsError`. Missing tools also do not prevent planning—they fail later when the detector chain is resolved. Consequently the documented `<detector> not ready (...)` reason cannot appear in a real scan; the tests only call `readinessSkipReason` directly. Exercise this end-to-end through `PlanSubprojects`, or hook the diagnostic into the all-detectors-not-ready resolution error instead.

bomly-guy (commented):
One advertised diagnostic is unreachable from the path that emits it.

internal/cli/opts/planning_diagnostics.go
bomly-guy: Agreed — fixed in 0ea726e. `readinessSkipReason` is gone from the discovery probe (planning never probes readiness, so the reason was unreachable from the path that emits it), and the diagnostic now lives on the resolution path instead:

- `resolveDetector` marks readiness failures with a typed `detectorNotReadyError`.
- When *every* link in a chain failed that way, `resolveDetectors` returns `noUsableDetectorError` — one line naming each link and what it needs, with the joined error preserved for `errors.Is`/`As`.
- A chain where a detector actually ran and failed keeps its verbatim error; the summary is scoped to readiness only.

Verified end to end with `mvn` off `PATH`:

```txt
Error: dependency resolution: subproject . (maven/maven): no usable detector: maven-detector not ready (mvn executable not found on PATH)
```

New tests in `internal/engine/pipeline_test.go`: the fallback-chain summary (both links named) and the negative case (a run failure stays verbatim).

bomly-guy:
@coderabbitai — all three nitpicks point at the same two branches in `detectorSkipReason`. I checked reachability before writing tests or docs for them, and neither can fire in a real scan:

- The probe surfaces candidates from `registry.DetectPackageManagers`, i.e. the **built-in catalog only**, and `detectorNamesForPackageManager` seeds its list from that same catalog. Every catalog manager with manifest evidence ships a detector chain, so `candidates` is never empty → `no detector registered for <pm>` is dead.
- With a non-empty chain, `plannedSubprojectForPackageManager` only fails on inputs the catalog cannot produce here (evidence that does not resolve, a relative execution-target location that never reaches this path) → `no resolvable <pm> manifest in this directory` is dead too.

The earlier human review on this PR rejected exactly this shape — an advertised diagnostic that the emitting path cannot reach — so rather than adding tests and a docs row for reasons users cannot hit, both branches are removed in 835a6f7. What remains is the one reachable message (`detector filter excludes every <pm> detector (...)`), which is already covered by `TestDiscoveryProbeExplainsDetectorFilterSkip`.

The invariant that message now leans on is pinned explicitly: `registry.TestEveryDetectablePackageManagerHasADetectorChain` fails if an evidence-only catalog entry is ever added, which is the condition that would make "the filter emptied it" a lie. The `no detector registered for <pm>` row is dropped from the `docs/SCAN_TARGETS.md` table since it can no longer be emitted.

PR #334, opened 2026-07-25
Sanitize subprocess diagnostics

bomly-guy:
> [!IMPORTANT]
> This PR is stacked on #327 so both endpoint and command logging use one fail-closed URL sanitizer. Merge #327 first, then rebase this branch onto `main` before merging #334.

## What changed

- add one shared sanitizer for recognized credential flags, npm registry credentials, and URL user information
- reuse the same fail-closed URL sanitizer for endpoint, Git target, and command logging
- preserve the Git error chain while keeping credential-bearing clone URLs out of errors
- log executable, sanitized arguments, and working directory for built-in install-first runners, native SwiftPM/pub/sbt, Git, external Syft/Grype, Java readiness, Pipenv environment checks, and managed plugin startup
- remove raw install arguments from engine orchestration logs
- stream raw subprocess stderr only at DEBUG verbosity (`-vv`)
- keep subprocess stderr out of default and INFO output and out of structured results
- pass the existing protocol-v1 `allowStdErrLogging` request signal when debug output is enabled
- report stderr byte counts in failures even when raw diagnostics are hidden

## Why

Command arguments and repository URLs can contain credentials, so Bomly sanitizes recognized secret-bearing values before logging them. Raw tool stderr is different: it is essential for diagnosing package-manager, analyzer, matcher, Git, Java, and plugin failures, but cannot be reliably sanitized without damaging the diagnostic.

The selected boundary makes raw stderr an explicit DEBUG feature. Users who select `-vv` receive the invoked tools diagnostics and must treat the resulting log as sensitive data.

The argument and URL sanitizer remains deliberately bounded. It recognizes documented credential-shaped flags and URL user information; it does not claim to classify arbitrary argument text, URL query values, or raw tool output.

## User impact

Default and INFO output remain unchanged. With `-vv`, users see the executable, sanitized arguments, working directory, and raw subprocess diagnostics. JSON and other structured outputs never contain raw subprocess stderr.

## Validation

- focused logging, detector, analyzer, Git, registry, plugin, CLI, and engine tests
- debug-visible and non-debug-hidden subprocess stderr regressions
- Java and Git diagnostic regressions
- external Syft and Grype build-tag build
- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`


bomly-guy:
Reviewed. This is the largest change in the batch and the parts I like are substantial: one shared `logging.SanitizeArgs` replacing the python-local `sanitizeCommand`, `CommandFields` giving every runner the same reproducible executable/args/working-dir triple, plugging the `resolved.URL` and git clone `source` leaks that [#327](https://github.com/bomly-dev/bomly-cli/pull/327) left open, dropping `install_args` from the engine log, and disabling go-plugin's stderr logger. Those are real fixes.

My concern is the blanket removal of subprocess stderr, which I think overshoots and takes several diagnostics down with it.

**1. `CommandStderr` no longer mirrors at `-v`, and the plumbing for it is now dead rather than removed.**

```go
func NewCommandStderr(_ io.Writer, _ bool) *CommandStderr {
	return &CommandStderr{}
}
```

Both parameters are ignored, but ~34 call sites still compute and pass `req.Stderr, req.Verbose`, and `sdk.DetectionRequest.Stderr` / `.Verbose` now have no consumer (`opts/options.go:149` is the only remaining writer of `Verbose`). Leaving a two-argument constructor that discards both arguments is an invitation for someone to "fix" it back six months from now. If the removal stands, delete the parameters and the `DetectionRequest` fields in the same PR so the intent is enforced by the compiler.

**2. The debuggability loss is real, and it hits the two most common failure modes.**

[docs/TROUBLESHOOTING.md:150](docs/TROUBLESHOOTING.md:150) currently promises that `-vv` output "is usually enough to file a useful bug report". After this change:

- `java -version` output is gone. `TestMavenDetectorReadyRequiresJava` / `TestDetectorReadyRequiresJava` / `TestNativeDetectorReadyRequiresJava` are all inverted to assert that `"Unable to locate a Java Runtime"` is *absent*, replaced by `"java runtime is unavailable: exit status 1 (diagnostic bytes: 82)"`. Missing/misconfigured Java is the single most common setup failure for the Maven/Gradle/sbt detectors, and `java -version` emits a fixed-format version banner — there is no credential surface here at all.
- `runGit` no longer includes stderr, so `verify --base <ref>` loses `fatal: bad revision`, and `git rev-parse` / `git diff` are local operations that cannot carry a token.

A per-command policy would keep the security property and the diagnostics: retain stderr for commands whose output we know the shape of (`java -version`) and for local, non-network git subcommands (`rev-parse`, `diff`, `checkout`), and drop it only for network-touching git (`clone`, `fetch`) and for arbitrary third-party build tools. Alternatively, keep mirroring behind a redacting line filter, or behind an explicit opt-in (`--show-tool-output`) defaulting off. Note the trust framing too: for a user running Bomly in their own terminal, a build tool's stderr reaching their own stderr is not disclosure to a third party — the real risk is CI log capture, which an explicit off-by-default flag addresses precisely.

**3. Bug: the sanitized clone error breaks the error chain.**

```go
safeErr = fmt.Errorf("%s", strings.ReplaceAll(err.Error(), source, safeSource))
```

`%s` instead of `%w` flattens the error, so `errors.Is` / `errors.As` against the underlying `*exec.ExitError` stops working for any clone from a URL containing userinfo — and it only breaks on that path, so it's the kind of asymmetry that shows up much later. Also, now that `runGit` returns only `"%w (stderr bytes: %d)"`, `err.Error()` can no longer contain `source` at all, so the `ReplaceAll` is dead code. Both go away if you drop the rewriting and just wrap: `fmt.Errorf("clone git repository %q: %w", safeSource, err)`.

**4. `sensitiveFlag` gaps — and this sanitizer is now output-facing, not just log-facing.**

`internal/detectors/python/resolution.go` routes `ResolutionMetadata.InstallCommand` through `logging.SanitizeArgs`, and that field is emitted in scan JSON. So a false negative here is a secret in a committed artifact, not just in a debug log. Two concrete gaps:

- Missing names: `auth`, `authorization`, `bearer`, `pat`, `passphrase`, `pass`, `pwd`. `--header "Authorization: Bearer …"` and `--auth user:token` both pass through untouched today.
- `sensitiveFlag` returns early unless `value[0] == '-'`, so npm's registry-scoped form `//registry.npmjs.org/:_authToken=<token>` is not matched: the `=`-split path takes `flag = "//registry.npmjs.org/:_authToken"`, `sensitiveFlag` rejects it on the leading-`/` check, and the token is emitted verbatim. The same string *with* leading dashes is correctly redacted, which shows the name-matching itself works — it's only the dash gate that misses. Dropping the gate (or applying the name check to the whole argument regardless of prefix) fixes it. Worth a test case, since this is a form users really do pass.

Also `redactURLUserinfo` only handles the userinfo component: `https://host/path?access_token=…` survives. Probably acceptable, but it should be stated as a known limit next to the function rather than implied to be general URL redaction.

**Smaller notes**

- `CommandFields` does not sanitize `executable`. Fine today (all callers pass a looked-up binary path), but a comment saying args are sanitized and the executable is assumed safe would prevent someone passing a full command string later.
- `TestSanitizeArgsDoesNotTreatOrdinaryAuthoredFlagsAsCredentials` is a good addition — false-positive coverage on a redactor is the half people skip. Consider adding `--sort-key` / `--ssh-key` to it, since `key` is in the sensitive list and both are plausible non-secret flags whose values would be silently redacted.

bomly-guy:
Re-reviewed at `Reuse fail-closed endpoint sanitization`. Three of my four points are resolved, and cleanly:

- The `%s`-instead-of-`%w` clone bug is gone, and `TestCloneIntoRedactsUserinfoAndPreservesExitError` pins *both* halves (userinfo removed, `*exec.ExitError` still unwrappable). That's the right test.
- `SanitizeArgs` now covers `auth`/`authorization`/`bearer`/`pat`/`passphrase`/`pass`/`pwd`/`header`, `redactURLUserinfo` and `SanitizeURL` fail closed on parse error, and narrowing `key` to `len(parts) == 1` fixes the `--ssh-key`/`--sort-key` false positive I raised as a nit. Folding `endpointForLog` from [#327](https://github.com/bomly-dev/bomly-cli/pull/327) into `SanitizeURL` while keeping the fail-closed behavior is a better outcome than either PR had alone.
- Keeping `DetectionRequest.Stderr`/`Verbose` is fair — removing exported `sdk` fields is a breaking change for the plugin repos that pin the SDK, and the doc comments now make the intent explicit rather than leaving them looking vestigial. Withdrawn.

**New issue introduced by the point-4 fix.** Dropping the `value[0] != '-'` gate from `sensitiveFlag` fixed the npm `//registry.npmjs.org/:_authToken=` case, but it also applies to the bare-argument branch:

```go
if sensitiveFlag(argument) {
    sanitized[index] = argument
    redactNext = true
    continue
}
```

Any *positional* argument that folds to one of those words now silently redacts the argument after it — a package name like `token` or `pass`, a Gradle task, a subcommand. `pip install pass some-other-pkg` would emit `pass [REDACTED]`. That lands in `ResolutionMetadata.InstallCommand` in scan JSON, so it's a visible wrong value in a user-facing artifact, not just a noisy log.

The gate is only load-bearing for this branch, because that's the one where a leading dash is what distinguishes "flag whose value follows" from "value". The `strings.Cut` branch above doesn't need it — that's where the `//registry:_authToken=` form is handled, and it already works without the gate. So restoring the dash check in the bare-argument branch only keeps both fixes:

```go
if strings.HasPrefix(argument, "-") && sensitiveFlag(argument) {
```

Worth a test asserting a positional `pass`/`token` leaves the following argument intact, alongside the existing `//registry:_authToken=` case.

**Still outstanding: point 2.** The docs were updated to state that raw stderr is never printed, which is honest, but the substantive argument wasn't addressed: `java -version` and local git subcommands are not arbitrary third-party output.

- [internal/detectors/java_ready.go](internal/detectors/java_ready.go) still discards the probe output, so a missing or broken JDK reports `java runtime is unavailable: exit status 1 (diagnostic bytes: 82)`. `java -version` emits a fixed-format version banner — there is no credential surface, and this is the most common Maven/Gradle/sbt setup failure by a wide margin.
- `runGit` still drops stderr for `rev-parse` and `diff`, which are local operations that cannot carry a token, so `--base`/`--head` typos lose `fatal: bad revision`.

Both are cases where the shape of the output is known to us, unlike an arbitrary build tool's stderr. A `retainStderr bool` on the counter (or a separate `system.CommandDiagnostics` for known-safe probes) would let those two keep their diagnostics without weakening the general rule. If you'd rather ship as-is, that's your call — but please record it as a deliberate diagnosability trade-off in the [#337](https://github.com/bomly-dev/bomly-cli/pull/337) residual-risk column, since "no Java" and "bad revision" becoming opaque is the change users will actually notice.

bomly-guy:
Re-reviewed at `fix(logging): preserve positional subprocess arguments`. The positional-argument regression is fixed with the gate restored to exactly the one branch that needs it, and `TestSanitizeArgsDoesNotTreatPositionalArgumentsAsCredentialFlags` pins it. I checked that the fix didn't cost the coverage it was added for: `//registry.npmjs.org/:_authToken=` is still asserted redacted at index 20 of `TestSanitizeArgsRedactsCredentialValuesAndURLUserinfo`, since that form goes through the `strings.Cut` branch. Ran `go test ./internal/logging ./internal/git ./internal/registry` against this head — all pass.

On the Java/local-Git stderr point: you've kept the uniform rule and recorded the cost in [#337](https://github.com/bomly-dev/bomly-cli/pull/337) — both the assurance inventory and [docs/SECURITY.md](docs/SECURITY.md) now name Java readiness checks and local Git revision/diff commands specifically, and state that failures keep the exit status and byte count but lose the tool's message, "in exchange for one consistent subprocess-output boundary". That's a legitimate call and it's now documented where a user hitting it will find it, so I'm dropping the point. One consistent boundary is easier to reason about than a per-command allowlist that someone has to keep correct.

Nothing outstanding from me on this PR.

bomly-guy:
Reviewed `feat(logging): show subprocess stderr in debug mode`. This walks back the uniform-drop decision I had accepted, and I think the new position is the better one — it's essentially the explicit-opt-in middle ground from my original review, landed properly rather than as a per-command allowlist:

- Default and `-v` stay secret-safe; `-vv` becomes the documented opt-in, with [docs/TROUBLESHOOTING.md](docs/TROUBLESHOOTING.md) and `dev-docs/ARCHITECTURE.md` both stating plainly that debug logs must be treated as sensitive because arbitrary tool output can't be reliably sanitized. That's the honest framing — a warning, not a false guarantee.
- The gate is central: `PipelineRequest` nils out `Stderr` unless verbosity ≥ 2, so detectors/matchers/analyzers can't accidentally mirror below debug, and `TestPipelineRequestExposesSubprocessStderrOnlyAtDebug` pins both directions. The `req.Stderr != nil` pattern in grype/govulncheck is sound *because* of that single gate.
- The diagnostics I flagged as lost are back where they matter: `-vv` recovers `Unable to locate a Java Runtime` (asserted end-to-end in `TestRoot_ScanCommand_MavenMissingJavaReturnsResolutionFailure`) and git's `fatal:` messages (`TestRunGitLogsStderrAtDebug`). Error *text* stays byte-count-only at every verbosity, so the redaction contract on returned errors — including the clone userinfo test — is untouched.
- [#337](https://github.com/bomly-dev/bomly-cli/pull/337) was updated in the same push, so the assurance inventory and `docs/SECURITY.md` describe the new boundary rather than the old one. I checked; no staleness.

Verified locally: build plus `go test` over logging, git, opts, node, python, govulncheck, engine — all pass. CI green.

**One gate bypass that needs the one-token fix:** [internal/cli/benchmark_cmd.go:65](internal/cli/benchmark_cmd.go:65)

```go
NativeScan: benchmarkNativeScanner(logger, streams.notificationWriter(), current.Verbosity > 0),
```

`benchmarkNativeScanner` builds its `PipelineRequest` directly ([benchmark_run.go:70](internal/cli/benchmark_run.go:70)), so it doesn't pass through the `Options.PipelineRequest` gate — and it still uses the old `> 0` threshold with a live writer. Net effect: the hidden benchmark command mirrors raw subprocess stderr at `-v`, contradicting the new documented rule ("hidden below debug verbosity") everywhere else. It's a hidden local command, so severity is low, but it's exactly the second-constructor drift the central gate exists to prevent. `current.Verbosity >= 2` fixes it; a comment on `benchmarkNativeScanner` noting it must mirror the opts gate would keep the invariant visible.

**Two minor notes, neither blocking:**

- `pluginLogger` now returns the null logger for verbosity < 2, where `-v` previously got hclog Info. So plugin *lifecycle* logs (start/handshake/exit) disappear at `-v` along with the stderr. Deliberate per the new comment, and consistent — just noting it's a small `-v` behavior change beyond the stderr question, in case anyone relies on those lines.
- `runGit` and `JavaReady` now buffer full stderr unbounded before the Debug log. Fine in practice (clone is `--quiet`, `java -version` is a banner), not worth a cap unless something noisy joins `runGit` later.

With the benchmark threshold fixed I have nothing else on this PR — and for what it's worth, I prefer this resolution to the one I signed off on yesterday.

bomly-guy:
Pushed 4ad31fd closing the gate bypass from my review:

- `benchmarkNativeScanner` now applies the same subprocess-stderr gate as `Options.PipelineRequest` — raw tool output is forwarded only at `-vv`. The gate lives in a named helper (`benchmarkSubprocessStderr`) with a comment explaining why this second construction site must mirror the opts gate, so the invariant is visible where it could next be broken, and `TestBenchmarkSubprocessStderrForwardsOnlyAtDebug` pins both directions.
- [internal/cli/benchmark_cmd.go](internal/cli/benchmark_cmd.go) moves from `current.Verbosity > 0` to `>= 2`, and the scanner's `verbose` parameter is renamed `debug` to match the new semantics. The `io.Discard` fallback is gone — a nil `Stderr` is the normal below-debug representation now, same as the opts path.

Verified locally: gofmt clean, `go build ./...`, `go vet`, and the full `internal/cli` + `internal/cli/opts` suites pass.

The two minor notes from my review stand as-is by design: plugin lifecycle logs moving from `-v` to `-vv` is the deliberate consequence of the unified threshold, and the unbounded stderr buffers in `runGit`/`JavaReady` aren't worth a cap for `--quiet` git and a `java -version` banner. Nothing else outstanding from me on this PR.

PR #335, opened 2026-07-25
feat(engine): report CI-readiness hints for package-manager mismatches

### LINKED ISSUE #247: CI-readiness hints: detect package-manager/lockfile mismatches and install policy gates
Follow-up from #245 (deferred out of the MCP revamp in #246).

The #245 comment describes CI failures that were not vulnerabilities at all — e.g. a pnpm `minimumReleaseAge` policy gate, or a lockfile generated with pnpm 11 while CI uses pnpm 9. Bomly could detect and report these as diagnostics so an agent (or human) learns "CI will fail even if the vulnerability is fixed" before pushing.

Candidate detections:
- package-manager version mismatch (lockfile format version vs the tool on PATH / engines field / packageManager field)
- lockfile format mismatch (e.g. lockfileVersion newer than the manager understands)
- install policy gates such as pnpm `minimumReleaseAge` that will reject a freshly published fix version

Surface: the diagnostics channel added in #246 (`Diagnostic{Stage, Source, Message}` on MCP responses; PipelineWarning on the CLI side).

bomly-guy:
Closes #247.

## What

Some CI failures are not vulnerabilities at all: a lockfile written by pnpm 11 against a CI that runs pnpm 9, a Corepack pin that disagrees with the committed lockfile, or a pnpm `minimumReleaseAge` gate that rejects the fix version because it was published an hour ago. Bomly now detects that class of problem and reports it as **CI-readiness hints**, so an agent (or a human) learns "CI will fail even if the vulnerability is fixed" before pushing.

New package `internal/engine/ciready` runs right after detection (`Pipeline.runCIReadiness`) and emits `PipelineResult.CIWarnings`.

## Checks (Node: npm, pnpm, Yarn, Bun)

| Check | Example |
| --- | --- |
| Lockfile format newer than the manager on PATH | `pnpm-lock.yaml` is 9.0, pnpm 8 installed — CI cannot read it |
| Lockfile format older than the manager on PATH | `yarn.lock` v1 with Yarn 4 installed — it migrates the file, so a frozen-lockfile step fails |
| Corepack pin vs. PATH | `packageManager: "pnpm@11.0.0"` while pnpm 9 is installed |
| Corepack pin vs. lockfile | `packageManager: "yarn@4.1.0"` next to `pnpm-lock.yaml` |
| `engines` vs. PATH | `engines.node: ">=22"` while Node 20 is installed |
| Install age/date gates | `minimumReleaseAge` (pnpm-workspace.yaml), `minimum-release-age` and `before` (.npmrc) |

Comparisons are by major version, so routine patch drift stays silent, and the pin check goes quiet under Corepack (where `pnpm --version` already reports the pinned version).

## Surfaces

- **CLI**: ⚠ children under the detection step for `scan` / `explain` / `diff`; warning logs at `-v` (the channel CI runs without a TTY see), probe detail at `-vv`.
- **MCP**: `Diagnostic{Stage: "ci-readiness", Source, Message}` in the `diagnostics` array of every tool response — the channel #246 added.

## Boundaries

- Read-only and network-free: reads manifests, lockfiles, `pnpm-workspace.yaml`, `.npmrc`, and runs `<manager> --version` for managers already on PATH (one probe per manager per run, cached).
- Never creates findings, never affects the exit code, never fails the pipeline. Unreadable or unparseable inputs are skipped silently — malformed lockfiles are the detectors' report to make.
- Deliberately excluded from `baselineMutationWarningCount`: a manager mismatch is not degraded scan coverage and must not block a baseline write.
- No new flag: the check is cheap, and gating it would hide it from exactly the CI runs that need it.

## Testing

- `make test`, `make lint`, `make generate` — all clean.
- New unit tests in `internal/engine/ciready` cover each check, both mismatch directions, dedup across subprojects, single-probe caching, container/non-Node/file targets, and malformed inputs (version probes are stubbed, so no PATH dependence).
- Pipeline test asserts `CIWarnings` are populated and are *not* reported as detector warnings; CLI test asserts the `ci-readiness` MCP stage tag.
- Verified end-to-end against a scratch project with a 6.0 lockfile, an `engines.node` constraint, and a `minimumReleaseAge` gate — all three hints fire with the machine's real pnpm 11 / Node 22.

No smoke test: hints go to the progress/diagnostics channel, not to the CLI JSON documents the goldens encode.

Docs: new `docs/CI_READINESS.md` (+ manifest/README/MCP links), pipeline stage added to both architecture overviews, decision-log and network-behavior entries in `dev-docs/ARCHITECTURE.md`, and the architecture tree/pipeline order updated in `CLAUDE.md` and `AGENTS.md`.


bomly-guy (commented):
Two correctness issues in the version probe need addressing.

internal/engine/ciready/ciready.go
```diff
@@ -0,0 +1,184 @@
...
+	return filepath.Dir(location)
+}
+
+// version returns the version reported by binary, probing PATH at most once per
+// binary per Inspect call.
+func (r *run) version(ctx context.Context, binary string) (string, bool) {
+	if cached, ok := r.versions[binary]; ok {
+		return cached.version, cached.ok
+	}
+	version, ok := r.toolVersion(ctx, binary)
```
bomly-guy: [P1] Make the version probe project-scoped. Corepack selects the package-manager version from the nearest `package.json` relative to the command's working directory, but this callback receives no directory and the result is cached once per binary. As a result, scans of another path/remote clone use Bomly's process cwd, and monorepo subprojects with different pins all reuse the first version, producing false pin/lockfile mismatches. Pass the inspected directory through, set `cmd.Dir`, and key any Corepack-sensitive cache by directory (or explicitly detect Corepack instead of treating it as a global PATH version).

internal/engine/ciready/ciready.go
```diff
@@ -0,0 +1,184 @@
...
+
+// commandVersion runs `<binary> --version` and returns the first version-like
+// token in its output.
+func commandVersion(ctx context.Context, binary string) (string, bool) {
+	if _, err := system.LookPath(binary); err != nil {
+		return "", false
+	}
+	ctx, cancel := context.WithTimeout(ctx, toolVersionTimeout)
+	defer cancel()
+	cmd := system.CommandContext(ctx, binary, "--version")
```
bomly-guy: [P1] This unconditional stage is not actually network-free. `pnpm`/`yarn` on PATH may be Corepack shims, and invoking `--version` downloads the project-pinned manager on demand when it is not cached, while also mutating Corepack's cache. That makes a plain scan contact a registry without `--enrich`, contrary to this PR's stated boundary and Bomly's explicit network rules. Corepack documents the on-demand download behavior: https://nodejs.org/download/release/v20.18.2/docs/api/corepack.html. Avoid executing a shim unless network is forcibly disabled and treat a cache miss as unknown, or otherwise move this behind an explicit network-enabled path.

bomly-guy:
Changed direction per review: the pipeline stage is gone (`internal/engine/ciready` deleted). Pushed as c42d8cc.

**Where it lives now.** The Node detectors find these while resolving and return them with the graph, as `sdk.ResolutionWarning{Code, Source, Message}` on `ManifestMetadata.Resolution` — next to the fallback provenance already recorded there.

On `Ready()`: it is the right home for "is this toolchain usable", but it cannot carry this. Its verdict is binary, and a non-nil error routes the subproject to `resolveFallback` — so reporting a lockfile-format mismatch through it would demote a working lockfile parser to Syft over an advisory-only condition. The plugin transport has the same gap: `ReadyResponse.Reason` is dropped when `Ready` is true. A detector that resolves a sound graph from a misconfigured project has found a *resolution* problem, so that is where it is reported.

**Both review comments are resolved by dropping the probe entirely.** No package manager is executed any more. Every check reads committed files only — `package.json`, the lockfile the detector already parsed, `pnpm-workspace.yaml`, `.npmrc` — and compares what the *repository* declares (`packageManager`, `engines`) against the committed lockfile format. That removes the Corepack on-demand-download hazard outright, and the directory-scoping bug goes with the per-binary version cache. It is also the more accurate comparison, since CI installs from the repository, not from the scanning machine's `PATH`. Accepted trade-off, documented: a project that pins nothing gets no version comparison.

**Every surface, since `-q` silences progress.** The warning rides the manifest, so it reaches:

- scan JSON at `manifests[].resolution.warnings`, with a stable `code` (`lockfile-format-mismatch`, `package-manager-mismatch`, `engines-constraint-mismatch`, `install-policy-gate`)
- the text and Markdown reports, above the summary
- ⚠ progress children, `-v` logs, and MCP diagnostics under the `detect` stage

`PipelineResult.ResolutionWarnings` stays separate from `DetectorWarnings` so a manager mismatch is not counted as degraded coverage and does not block a baseline write.

**Verification.** `make test`, `make lint`, `make generate` clean. New unit tests in `internal/detectors/node` for each check plus detector-level tests that the warnings land on the manifest; engine tests for collection, dedup, and the DetectorWarnings separation; output test that the JSON carries code/source; render tests for the report notices. Verified end-to-end on a scratch pnpm project across all four surfaces.

No golden churn: none of the pinned smoke fixtures declare `packageManager` or `engines` (except `example-javascript-bun`, whose pin matches its own lockfile), and the one `pnpm-workspace.yaml` has no `minimumReleaseAge`, so no manifest gains a `resolution` block.

internal/detectors/node/resolution_warnings.go
```diff
@@ -0,0 +1,394 @@
...
+	var warnings []sdk.ResolutionWarning
+	source := managerBinaries[manager]
+	if source == "" {
+		source = "npm"
+	}
+	if minutes, ok := workspaceMinimumReleaseAge(projectDir); ok {
+		warnings = append(warnings, releaseAgeWarning(source, "pnpm-workspace.yaml sets minimumReleaseAge", minutes))
+	}
+	npmrc := readNpmrc(projectDir)
+	if minutes, ok := npmrcMinutes(npmrc, "minimum-release-age"); ok {
```
bomly-guy: [P1] Parse install gates according to the effective manager and its version. This currently applies the same `.npmrc` keys and minute units to npm, pnpm, Yarn, and Bun. npm's actual setting is `min-release-age` (not `minimum-release-age`) and its unit is days, while `before` is npm-specific: https://docs.npmjs.com/cli/install/#min-release-age. pnpm 10 accepts `minimum-release-age` in `.npmrc` as minutes, but pnpm 11 reads only auth/registry settings there and puts `minimumReleaseAge` in `pnpm-workspace.yaml`: https://pnpm.io/settings#minimumreleaseage. As written, real npm gates are missed, pnpm-11/Yarn/Bun projects can get false warnings for ignored keys, and the displayed duration can be wrong by 1440x. Determine the effective manager from `packageManager` (falling back to the detector manager) and handle each manager/version's key and units explicitly.

internal/detectors/node/resolution_warnings.go
```diff
@@ -0,0 +1,394 @@
...
+	var warnings []sdk.ResolutionWarning
+	if pinnedManager != manager {
+		if committed := committedLockfile(projectDir, manager, lockfile); committed != "" {
+			warnings = append(warnings, sdk.ResolutionWarning{
+				Code:   sdk.ResolutionWarningPackageManager,
+				Source: binary,
+				Message: fmt.Sprintf("package.json pins packageManager %q but the project commits %s; CI installs with the pinned manager and will not use this lockfile",
+					collapse(pkg.PackageManager), committed),
+			})
+		}
```
bomly-guy: [P2] Account for lockfile interoperability before saying the pinned manager will not use this file. npm explicitly accepts `yarn.lock` as install input when no shrinkwrap/package-lock is present (see the precedence list in https://docs.npmjs.com/cli/install/), so `packageManager: "npm@..."` plus `yarn.lock` currently produces a false `package-manager-mismatch` claiming npm will ignore it. Other managers also import or migrate foreign lockfiles. Model accepted/imported lockfiles separately from unsupported ones, and only emit the hard mismatch when the pinned manager/frozen workflow actually cannot consume the committed file.

internal/cli/render/scan.go
```diff
@@ -127,6 +128,80 @@ func Scan(g *sdk.Graph, registry *sdk.PackageRegistry, findings []sdk.Finding, m
...
+	index := make(map[string]int)
+	for _, m := range manifests {
+		if m.Resolution == nil {
+			continue
+		}
+		for _, warning := range m.Resolution.Warnings {
+			// Messages embed values read from scanned repository content
+			// (version pins, config values); collapse embedded newlines so a
+			// crafted value cannot inject extra lines into rendered output.
+			message := collapseWhitespace(warning.Message)
```
bomly-guy: [P1] Sanitize terminal control sequences, not only whitespace. Both `warning.Message` and `m.Path` contain repository-controlled data; `collapseWhitespace` leaves CSI/OSC escapes and other control bytes intact, and `Scan` later places the notice inside `Style(...)`. A malicious repository can therefore embed an ESC + `[2J` sequence in `package.json`/`.npmrc` (or an escape in a filename) to clear or forge terminal output. Apply the CLI's untrusted-text sanitizer (covering CSI, OSC, and C0/DEL controls) before rendering both fields, and add a text-output test with encoded escape sequences.

internal/detectors/node/resolution_warnings.go
```diff
@@ -0,0 +1,394 @@
...
+	if err := yaml.Unmarshal(data, &workspace); err != nil {
+		return 0, false
+	}
+	if workspace.MinimumReleaseAge == nil || *workspace.MinimumReleaseAge <= 0 {
+		return 0, false
+	}
+	return *workspace.MinimumReleaseAge, true
+}
+
+// readNpmrc parses dir/.npmrc into lowercase keys. Values are returned verbatim.
```
bomly-guy: [P2] Add the required fuzz coverage for this new untrusted-repository parser. The existing npm/pnpm/Yarn lockfile fuzz targets stop at graph parsing; they do not exercise `ResolutionWarnings`' package.json/YAML/.npmrc conversion or `yarnLockfileFormat`. The repository's parser policy requires bounded valid/malformed/truncated seeds, deterministic success/failure, and registration in `scripts/run-fuzz.sh` for materially changed in-process parsers. A single project-config fuzz target can cover these warning parsers, or the exclusion and reason need to be recorded in `test/assurance/PARSER_FUZZING.md`.

bomly-guy (commented):
Re-review of the latest head: the previous Corepack/PATH issues are resolved by deleting the probe. The detector-owned replacement has four remaining issues.

internal/detectors/node/resolution_warnings.go
bomly-guy: Fixed in 55904ac. Gates are now keyed to the manager that will actually run the install — the `packageManager` pin when the project declares one, otherwise the detector's manager — and each key/unit is handled explicitly against its own documentation:

- pnpm: `minimumReleaseAge` (minutes) from `pnpm-workspace.yaml` only. Current pnpm reads just auth and registry settings from `.npmrc`, so `minimum-release-age` there is no longer treated as an active gate. pnpm-10 repos that set it in `.npmrc` therefore get no warning; version-sniffing the cutover felt worse than a documented gap.
- npm: `min-release-age` (days) and `before` (a date) from `.npmrc`. The 1440x duration error is gone — `min-release-age=7` renders as "7 days".
- Yarn and Bun: no committed gate modelled, so no warnings.

Covered by TestPackageManagerWarnings_NpmInstallGatesUseDays and TestPackageManagerWarnings_GatesAreScopedToTheEffectiveManager (which asserts silence for keys pnpm ignores, and that an npm pin suppresses pnpm's gate).

internal/detectors/node/resolution_warnings.go
bomly-guy: Fixed in 55904ac. There are now two documented tables instead of one assumption: `consumedLockfiles` (what a manager reads as install input) and `unreadLockfiles` (what it is documented not to read). A warning fires only for the second, so anything neither table covers stays silent.

Verified against the managers' own docs: npm drives installs from package-lock.json, npm-shrinkwrap.json, or yarn.lock, so your example no longer warns; Bun converts pnpm-lock.yaml on install; pnpm's conversion is the separate manual `pnpm import`, so a pnpm pin beside only a yarn.lock is still a real mismatch. Bun's handling of npm/Yarn lockfiles is undocumented either way, so it is deliberately left out of both tables. The code is `lockfile-unsupported` now rather than `package-manager-mismatch`, since that is what it actually asserts. Tests: TestPackageManagerWarnings_NpmAcceptsYarnLockfile, _UndocumentedCombinationsStaySilent, _PinnedManagerDoesNotReadCommittedLockfile, and a table test for lockfileUnread.

internal/cli/render/scan.go
bomly-guy: Fixed in 55904ac. Added `render.SanitizeUntrusted`, which drops every ESC-introduced sequence — CSI, the OSC/DCS/SOS/PM/APC string family up to BEL or ST, and charset-designator style escapes — plus any remaining C0 control byte and DEL, then folds whitespace. Both the message and the file name go through it in `WarningNotices`, and the progress-child builder uses it too.

Tests: TestSanitizeUntrusted covers each sequence family, TestWarningNotices_StripsTerminalControlSequences feeds `\x1b[2J`, an OSC title with BEL, and a CRLF forgery through the notice builder, and TestScanRendersWarningNoticesWithoutControlSequences asserts the rendered `Scan` output for encoded escapes. `collapseWhitespace` is gone.

internal/detectors/node/resolution_warnings.go
bomly-guy: Added in 55904ac. `FuzzPackageManagerWarnings` writes each fuzz input to package.json, pnpm-workspace.yaml, .npmrc, and yarn.lock in a temp project and runs the checks for all four managers, asserting no panic, a deterministic warning count for a fixed input, and single-line messages. Seeds cover valid minimal documents, a truncated JSON object, an empty pin, control bytes, and the empty string, all bounded by nodetest.MaxFuzzInputSize.

Registered in `scripts/run-fuzz.sh` and listed in `test/assurance/PARSER_FUZZING.md` under a new "Node project configuration" row. `yarnLockfileFormat` reads only the head of yarn.lock and shares the YAML path the target exercises.

internal/engine/ciready/ciready.go
bomly-guy: Obsolete: `internal/engine/ciready` was deleted in c42d8cc. No version probe remains — the checks read committed files only, so there is no directory-scoped cache to get wrong. The per-subproject working directory now comes from the detector's own request.

internal/engine/ciready/ciready.go
bomly-guy: Obsolete, and the underlying point drove the redesign: `internal/engine/ciready` was deleted in c42d8cc and nothing executes a package manager any more. Corepack shim downloads were exactly why — the checks now compare what the repository declares (`packageManager`, `engines`, committed lockfile, install-gate config), which is also what CI installs with.

bomly-guy:
Filed https://github.com/bomly-dev/bomly-cli/issues/340 for the remaining stages: match/analyze/audit warnings are still untyped `PipelineWarning`s that reach only progress and MCP, so they stay invisible in JSON and under `-q`, and baseline gating counts them unconditionally. This PR's scope stays detection-only.

PR #337, opened 2026-07-25
Document security and trust boundaries

bomly-guy:
> [!IMPORTANT]
> This is the final issue 026 assurance PR and must merge last. The documentation describes the combined intended state after every dependency below has merged.

## Dependencies

- #325 — proxy fallback honors Bomly no-proxy settings
- #326 — filesystem, archive, cache, Git, and output containment
- #327 — endpoint credential redaction
- #328 — execution, plugin environment, MCP authority, and read-only remediation assurance
- #329 — proxy, CA, redirect, private-endpoint, and HTTP error boundaries
- #330 — automatic baseline path containment
- #331 — stable, secret-safe MCP errors
- #332 — plugin download, extraction, and metadata resource limits
- #334 — sanitized command logging with raw subprocess diagnostics at DEBUG
- #336 — configuration, baseline, SBOM, and matcher response limits

PRs #320 and #324 are already on `main` and supply the repository-configuration and matcher-intent foundations.

## What changed

- Add a concise public guide explaining safe defaults, network authority, file and resource limits, plugin trust, MCP behavior, and vulnerability reporting.
- Explain that `-vv` includes raw subprocess stderr for troubleshooting and that debug logs must be handled as sensitive data.
- Add a maintainer assurance inventory mapping each trust boundary to its control, regression evidence, delegated authority, and residual risk.
- Record the CI workflow permission and credential inventory.
- Record the final issue 026 evidence index and acceptance disposition.
- Link the new pages from the documentation index, architecture overview, and CI guide.

## Why

The controls are intentionally split across small production and assurance PRs. This final document makes their combined security model reviewable in one place and states what Bomly does not guarantee: private endpoints and DNS rebinding remain supported through trusted configuration, enabled native plugins have user privileges, debug subprocess output may contain sensitive tool diagnostics, explicit Git URL scans have no hard byte or disk cap, host tools own registry credentials and network behavior, ordinary project evidence has no global byte limit, and local cache files are trusted state without a separate read-size cap.

## Validation

- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`

All passed locally. Generated documentation produced no unrelated drift.


bomly-guy:
Reviewed. This is the right document to have, and it does the hard part well: separating what Bomly guarantees from authority the user deliberately delegates, with a named residual-risk column instead of blank cells. The "Authority Model" table — each user choice mapped to exactly what it grants — is the thing a security reviewer actually wants, and the explicit statement that automatic baseline discovery survives *because* its only power is changing a finding's policy status is the correct justification rather than a hand-wave.

I verified every test name cited in the Boundary Inventory. All 14 I sampled exist and are in the packages implied (`TestPipelineRequiresExplicitMatcherIntent`, `TestPlanSubprojectsRecursiveDoesNotFollowSymlinkedDirs`, `FuzzLoad`, `FuzzLoadFile`, `FuzzUnmarshalAutoJSON`, `TestCommandContextResolveExecutionTarget_Image`, `TestDetectorApplicable_ContainerTarget`, `TestSyftSourceInput_UsesContainerReferenceForContainerTargets`, `TestExternalDetectorProvidesAdvertisedRemediationHints`, `TestDerivePackageRemediation`, `TestValidateHintsSanitizesAndBoundsAdvice`, `TestCollectHintsBoundsAndSanitizesDiagnostics`, `TestDeriveRejectsUnadvertisedAndUnknownHints`, `TestCommandContextInitialize_RequiresExplicitProjectConfig`). Naming tests rather than files is what makes this maintainable — worth propagating to the `test/assurance/*.md` docs in [#326](https://github.com/bomly-dev/bomly-cli/pull/326) and [#328](https://github.com/bomly-dev/bomly-cli/pull/328), which currently name files or nothing.

**One claim that won't be true as written:**

> [docs/SECURITY.md](docs/SECURITY.md): "Tool failures return stable error categories; detailed internal errors stay in local logs without credential values."

and the matching MCP row: "detailed internal errors stay in local logs". As implemented in [#331](https://github.com/bomly-dev/bomly-cli/pull/331), `toolErrorResult` logs only tool, category, and `%T` of the wrapper — the cause is discarded, at every verbosity. So there is no local log holding the detail. I've suggested on #331 that it log the cause at Debug, which would make this sentence accurate; if that isn't taken, this line needs rewording to "failures are reported as categories; stage-level detail is available only from components that log independently."

**Two smaller wording items:**

- "Atomic writers do not follow a final symlink where the contract rejects one" (Output paths row) is doing a lot of work with "where the contract rejects one". Either name the writers that reject it, or state the property unconditionally. As written a reader can't tell which case they're in, which defeats the purpose of the table.
- The Git targets row says "Git arguments and URLs are logged with credentials removed" and cites "command-sanitizer and Git credential tests" — the only row without concrete test names. Given the `SanitizeArgs` gaps I flagged on [#334](https://github.com/bomly-dev/bomly-cli/pull/334) (`auth`/`bearer`/`pat` not in the sensitive list; npm's `//registry:_authToken=` form unmatched), I'd soften this to "credential-shaped flag values and URL userinfo are removed" — the current phrasing reads as a completeness claim that the redactor doesn't support.

**Merge ordering** is already stated ("must merge after every listed production and test change") — good, and worth holding to strictly, since the Acceptance Disposition table reads as present-tense fact. Two of the listed PRs ([#332](https://github.com/bomly-dev/bomly-cli/pull/332)/[#326](https://github.com/bomly-dev/bomly-cli/pull/326)) currently conflict at the `extractZipArchive` signature, so the ordering isn't just cosmetic here.

Last thought: the Review Checklist at the end is the piece most likely to decay, because nothing enforces it. Consider linking it from the Feature Checklist in [CLAUDE.md](CLAUDE.md) / `AGENTS.md`, which is where someone adding a new input or subprocess is actually looking.

bomly-guy:
Re-reviewed at `Align the security assurance guide with review findings`. All of it addressed:

- The MCP row and [docs/SECURITY.md](docs/SECURITY.md) now say local logs hold the category and unwrapped cause *type*, never the cause text, which matches what [#331](https://github.com/bomly-dev/bomly-cli/pull/331) actually implements, and the generic-validation trade-off is recorded as a residual rather than left implicit.
- The stale `TestNewHTTPClientNoProxyAppliesToStandardProxyFallback` reference is updated to the renamed test from [#325](https://github.com/bomly-dev/bomly-cli/pull/325) — that rename happened *after* my review, so the inventory catching it is a good sign the citations are being maintained rather than written once.
- "Atomic writers do not follow a final symlink where the contract rejects one" is now split into what `baseline.WriteAtomic` guarantees versus what general output writes do, with the symlink-following behavior of the latter stated as a residual. That's the ambiguity gone.
- Git row softened to "recognized credential-shaped argument values" and now names concrete tests.
- HTTPS→HTTP redirect downgrade added to the HTTP residual column, matching [#329](https://github.com/bomly-dev/bomly-cli/pull/329)'s new "Intentional limits" text.
- The Review Checklist is linked from [CLAUDE.md:154](CLAUDE.md:154), so it's reachable from where someone adding a new input actually looks.

One thing to carry into the merge: I've flagged on [#334](https://github.com/bomly-dev/bomly-cli/pull/334) that the `java -version` probe and local `git rev-parse`/`diff` still discard their stderr, so "no Java" and "bad revision" are now opaque at every verbosity. If #334 ships as-is, the subprocess row's residual column should say so — right now it reads as a pure security win with no diagnosability cost. Also still worth confirming before this merges last: [#326](https://github.com/bomly-dev/bomly-cli/pull/326) and [#332](https://github.com/bomly-dev/bomly-cli/pull/332) currently produce a compile error in whichever merge result lands second (`extractZipArchive` arity), so the Acceptance Disposition table's present tense depends on that being sequenced.

bomly-guy:
Re-reviewed at `docs: record subprocess diagnostic tradeoff`. The last item is closed: the subprocess row and [docs/SECURITY.md](docs/SECURITY.md) now name Java readiness checks and local Git revision/diff commands explicitly, and state the cost rather than presenting the stderr rule as a pure win. "This loses some troubleshooting detail in exchange for one consistent subprocess-output boundary" is the right framing — it's the trade-off stated as a trade-off.

Also noting the inventory has kept pace with the other branches on its own: it picked up `TestResolveGitHubReleaseRejectsOversizedMetadata` from [#332](https://github.com/bomly-dev/bomly-cli/pull/332) and the new OSV/KEV/Scorecard limit tests from [#336](https://github.com/bomly-dev/bomly-cli/pull/336) without being asked. That's the property that makes this document worth maintaining.

The [#326](https://github.com/bomly-dev/bomly-cli/pull/326)/[#332](https://github.com/bomly-dev/bomly-cli/pull/332) compile conflict I flagged is also resolved — #326 is now stacked on #332, so the merge ordering the Acceptance Disposition table depends on is enforced by the branch graph rather than by convention. No concerns left from my side on any PR in this batch.

PR #338, opened 2026-07-25
fix(pip): scope direct dependencies, name project roots, diagnose old pip

### LINKED ISSUE #272: pip: requirements.txt project renders as "root" in the manifest tree
In the grouped manifest tree, every subproject renders a real project name except pip, which renders the literal string `root`:

```
  ├─ services/api (subproject, gomod)
  │  └─ acme.example/api — 2 packages [go.mod]
  └─ services/billing (subproject, pip)
     └─ root — 12 packages [requirements.txt]
```

**Repro:** `bomly scan --recursive` (bomly 0.18.0) on a monorepo fixture containing `services/billing/requirements.txt` (two pinned entries, no lockfile), alongside npm and Go subprojects.

**Expected:** something meaningful for a requirements.txt project — the subproject directory name (`billing`) seems like the natural fallback, matching how other detectors name their root when the manifest carries no name.

### LINKED ISSUE #273: pip: every installed distribution is reported as a direct dependency
With a `requirements.txt` containing exactly two entries (`flask==3.1.1`, `requests==2.32.4`) and no lockfile, the pip detector resolves 12 packages — correct — but reports **all 12 as direct**, and the Top-level dependencies table lists pure transitives (`blinker`, `certifi`, `charset-normalizer`, `click`, `idna`, `itsdangerous`, `jinja2`, `markupsafe`, `urllib3`, `werkzeug`) alongside the two requested packages.

Observed with bomly 0.18.0 (`bomly scan --recursive`): a polyglot fixture whose other ecosystems contribute 6 direct deps shows `18 direct` overall — the pip subgraph contributes 12/12 direct, 0 transitive.

The npm and Go detectors classify direct vs transitive correctly in the same scan.

**Likely cause:** the isolated-environment resolution path (`pip install` + `pip inspect --local`) reports every installed distribution without distinguishing requested packages from their dependencies. `pip inspect` output carries `requested: true/false` per package, which looks like enough signal to scope the graph.

**Expected:** only the packages listed in `requirements.txt` are direct; the rest are transitive with edges derived from each distribution's `Requires-Dist` metadata.

### LINKED ISSUE #274: pip: opaque failure when ambient pip lacks `pip inspect` (pip < 22.2)
On a machine where `python3` is the macOS system Python 3.9 (pip 21.2.4), the pip detector fails and degrades to syft with this warning:

```
⚠ pip-detector unavailable (pip detector: resolve isolated environment graph: run pip detector: exit status 1) — resolved services/billing/requirements.txt with syft-detector; transitive dependencies may be missing
```

The actual cause is that `pip inspect` only exists in pip ≥ 22.2, and the venv the detector creates inherits the ambient interpreter's pip. Nothing in the message points at the pip version, so the user has no path from "exit status 1" to "upgrade Python/pip". (Observed with bomly 0.18.0; with Python 3.12 on PATH the same scan resolves natively.)

Two improvements, either or both:

1. **Diagnose the known cause:** when `pip inspect` fails, check `pip --version` and, if it is < 22.2, say so in the warning and name the minimum version. Alternatively, upgrade pip inside the freshly created venv (`python -m pip install --upgrade pip`) before inspecting — the venv is already isolated and network-bound at that point.
2. **Document the requirement:** DETECTORS.md's prerequisites table says `pip inspect` "reads the local Python environment — None" for network/setup, but never states the minimum pip version.

bomly-guy:
Fixes #272, fixes #273, fixes #274.

## #273 — every installed distribution was reported as direct

`pip inspect` returns a flat installed set and has no `requested_by` field, so the `pkg.Requested || len(pkg.RequestedBy) == 0` condition was always true and every package got an edge from the root. A two-entry `requirements.txt` reported 12/12 direct.

Now the graph is built in three passes: `requires_dist` edges first, then the direct set (names the project's requirements files declare, plus the installer's `REQUESTED` marker for what those files cannot name — `-r` includes, environments populated by another front-end), then re-parenting anything left without a parent so the graph keeps a single root. The same repair runs after tool packages (`pip`, `setuptools`, …) are filtered out.

## #272 — requirements.txt projects rendered as `root`

The synthetic root is now named after `pyproject.toml`'s project name, else the subproject directory, else the scanned repository, else the project directory. Bomly's own `bomly-git-*` clone dirs are skipped — they are random per run, so remote scans would otherwise be non-deterministic and meaningless. Applies to the pip-inspect graph, `requirements.lock`, `Pipfile.lock`, and the Poetry fallback.

## #274 — opaque failure when ambient pip lacks `pip inspect`

`pip inspect` landed in pip 22.2, and `python -m venv` seeds the venv with the ambient interpreter's pip, so macOS system Python 3.9 (pip 21.2.4) produced a venv that could not inspect itself. Bomly now checks the venv's pip version and upgrades it in place — the environment is isolated and already network-bound for the install. If the upgrade cannot run, the detector fails with the pip version and the minimum named, which surfaces through the existing fallback notice instead of `exit status 1`.

Docs: pip and pipenv prerequisites now state the pip ≥ 22.2 requirement, and the `pip inspect` row in the network-behavior table no longer claims the class never reaches the network (it does, when populating the environment it inspects).

## Verification

This machine is the #274 repro (Python 3.9.6 / pip 21.2.4). Before, a `flask` + `requests` `requirements.txt` degraded to syft; now:

```
✓ 14 packages in 1 manifest   (2 direct, 12 transitive · runtime 14, dev 0)
  └─ services/billing (subproject, pip)
     └─ billing — 14 packages [requirements.txt]

Top-level dependencies
  NAME       VERSION   LICENSE   SCOPE     VULNS
  flask      3.1.1     -         runtime   -
  requests   2.32.4    -         runtime   -
```

`make test`, `make lint`, and `make generate` are clean. Unit tests cover the direct/transitive split (with and without `REQUESTED`), orphan re-parenting, root naming from each source, the pip upgrade, and the unsupported-version error.

Smoke goldens `scan-python-pip`, `scan-python-pip-reachability`, and `scan-python-pipenv` were regenerated locally; the only change in each is the root rename (`pkg:pypi/root` → the repository name). `scan-python-poetry` and `scan-python-uv` take the lockfile fast-path with named roots and are unchanged.

## Summary by CodeRabbit

* **Dependency Resolution**
  * Improved Python dependency graphs by accurately distinguishing direct and transitive dependencies.
  * Added clearer project root naming across Python package managers.
  * Orphaned dependencies are now safely attached to the project root.

* **Compatibility**
  * Added support for automatically upgrading outdated pip versions required for environment inspection.
  * Pipenv environments with older pip versions now fall back to lockfile parsing.

* **Documentation**
  * Clarified network behavior, pip requirements, dependency relationships, and upgrade failure handling.


internal/detectors/python/common.go
```diff
@@ -106,7 +111,11 @@ func (d baseDetector) resolveGraph(stderr io.Writer, projectPath string, verbose
 		return nil, fmt.Errorf("run %s: %w", detectorName, err)
 	}
 
-	depsGraph, err := depGraphFromPipInspect(out.Bytes())
+	declared, err := declaredPythonDependencies(cmd.Dir)
```
bomly-guy: [P1] `declaredPythonDependencies` is not safe to use as the direct-dependency set here. It feeds `poetry.lock` and `uv.lock` through `collectLoosePythonManifestDependencies`, which adds every `name = ...` package record, and similarly scans `Pipfile.lock` loosely. Since `resolveGraph` is shared by Poetry, UV, and Pipenv, their pip-inspect fallback will attach every locked transitive package to the root, recreating #273. Build this set only from actual direct declarations (`requirements*`, the dependency tables in `pyproject.toml`, and `Pipfile`); lockfile package records must not count as declarations.

internal/detectors/python/venv.go
```diff
@@ -94,3 +98,97 @@ func createPythonVenv(ctx context.Context, base baseDetector, req sdk.DetectionR
...
+		// inspect steps report whatever is actually wrong.
+		logger.Debug("could not read pip version from isolated environment", zap.String("detector", detectorName), zap.Error(err))
+		return nil
+	}
+	if pipSupportsInspect(version) {
+		logger.Debug("isolated environment pip supports inspect", zap.String("detector", detectorName), zap.String("pip_version", version))
+		return nil
+	}
+	logger.Info(fmt.Sprintf("%s upgrading pip in the isolated virtualenv (pip %s predates `pip inspect`, added in pip %s)", detectorName, version, minPipInspectLabel))
+	if err := base.install(ctx, req, detectorName+" (pip upgrade)", []string{venvPython, "-m", "pip", "install", "--upgrade", "pip"}); err != nil {
```
bomly-guy: [P1] Do not silently install/upgrade the package manager. The repository's non-negotiable contract says to assume package managers exist and not add PM installation logic; this also introduces an extra unpinned network mutation before resolution. Because `base.install` appends `req.InstallArgs`, arguments intended for the project install are also appended to `pip install --upgrade pip` and can make this bootstrap fail. Detect pip < 22.2 and return the actionable compatibility error (or use an already-compatible interpreter), but do not mutate pip.

bomly-guy (commented):
Two blocking correctness/behavior issues in the pip changes.

internal/detectors/python/common.go
bomly-guy: Agreed, and thanks — this would have recreated #273 for Poetry/uv/Pipenv.

Fixed in 6af600b: the direct set now comes from `directPythonDeclarations`, which reads only hand-authored declarations — `requirements.txt` / `-dev` / `.in`, the dependency tables in `pyproject.toml` (PEP 621 `[project]` + optional-dependencies, PEP 735 `[dependency-groups]`, `[tool.poetry]` dependencies / dev-dependencies / group.*.dependencies, `[tool.uv] dev-dependencies`), and the `Pipfile` `[packages]` / `[dev-packages]` tables. Parsed with the TOML decoder rather than the loose line scanner, and Poetry's `python` interpreter marker is dropped.

`requirements.lock` is excluded too, for the same reason you gave: it is pip-compile output carrying the full closure, and the inspect path is reachable when the lockfile fast-path fails to parse.

`declaredPythonDependencies` is untouched and still backs `filterPythonToolPackages` — that asks the looser "does this package belong to the project at all" question, where a lockfile record is legitimate evidence. `TestDirectPythonDeclarationsExcludesLockfiles` pins the split.

internal/detectors/python/venv.go
bomly-guy: You're right on all three counts — the `InstallArgs` contamination alone would have broken `--install-arg --require-hashes` on exactly the machines that hit this. Removed in 6af600b.

`ensurePipInspectSupport` is now `verifyPipInspectSupport`: it reads `pip --version` in the venv and returns the actionable compatibility error, nothing else. No install, no mutation. The check runs *before* the requirements install, so an uninspectable environment costs no network round trip; the test asserts the install log stays empty.

Behavior on the reporter's setup (this machine — Python 3.9.6 / pip 21.2.4):

```
⚠ pip-detector unavailable (pip detector: prepare isolated Python environment: isolated environment
has pip 21.2.4, but `pip inspect` requires pip 22.2 or newer; upgrade pip for that interpreter
(python -m pip install --upgrade pip) or put a newer Python on PATH) — resolved requirements.txt
with syft-detector; transitive dependencies may be missing
```

That is #274's option 1 — the fallback still yields a graph, and the notice now names the cause and the fix. Docs updated to say plainly that Bomly never installs or upgrades pip, and the architecture entry records why.

docs/detectors/ecosystems/python/pip.md
```diff
@@ -31,6 +31,10 @@ Bomly uses this chain when it finds `pip` evidence.
 
 The isolated virtualenv lives under the OS temp directory and is keyed by the project path. Bomly recreates it before installing so stale packages cannot leak into the scan, then inspects that venv directly. Ambient site-packages are never accepted as the project graph.
 
+Direct vs. transitive comes from the packages your requirements files name (plus the installer's `REQUESTED` marker); everything else in the environment is wired under its parents using each distribution's `Requires-Dist` metadata.
+
+The graph's root node is named after `pyproject.toml`'s project name when one is declared, and after the project directory otherwise — a `requirements.txt` project carries no name of its own.
+
```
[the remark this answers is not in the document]
bomly-guy: Fixed in 6af600b — the prose source (`internal/support/prose/detectors/pip.md`) now lists the full chain (pyproject name → subproject directory → repository name → project directory) and the `bomly-git-*` exclusion, regenerated via `make generate`.

internal/detectors/python/common.go
```diff
@@ -106,7 +111,11 @@ func (d baseDetector) resolveGraph(stderr io.Writer, projectPath string, verbose
...
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
bomly-guy: Fixed in 6af600b, wrapped as `collect declared dependencies for %s: %w`. (The call also moved to the new `directPythonDeclarations` collector.)

bomly-guy:
Pushed 6af600b addressing both P1s and the nitpicks.

**P1 — lockfile records counted as direct declarations.** Real, and it would have recreated #273 for Poetry/uv/Pipenv on their inspect path. The direct set now comes from a dedicated collector over hand-authored files only; `requirements.lock` is excluded on the same grounds. Details in the inline reply.

**P1 — pip self-upgrade.** Dropped entirely; the detector only diagnoses now, before the install runs. Details in the inline reply.

**Nitpicks:**
- `attachOrphansToRoot` now reattaches by root-reachability (BFS) instead of parent count, so a `requires_dist` cycle cannot strand a component. Covered by `TestDepGraphFromPipInspectAttachesCycles`.
- `readPipVersion` logs its subprocess at DEBUG with binary, args, and working dir, matching `createPythonVenv`.
- Docs: root-naming chain corrected; the pip network-behavior section now enumerates exactly which commands run in the temp venv and states that pip is never installed or upgraded.

`make test`, `make lint`, `make generate` clean. Re-ran the three python smoke cases against their pinned repos — all still match the goldens committed in the first commit, so no further regeneration.

PR #339, opened 2026-07-26
feat(mcp): add npm wrapper and MCP Registry metadata

bomly-guy:
Unblocks **L3-1** in the marketing listings tracker: publishing Bomly to the official [MCP Registry](https://registry.modelcontextprotocol.io).

## Why

The registry accepts only npm, PyPI, NuGet, Cargo, OCI, or MCPB packages. Bomly ships a Go binary, so there was nothing to point a registry entry at. The same gap blocks the community MCP directories (mcpservers.org, PulseMCP, Glama, MCP.so), which all expect a one-line `npx` install.

Of the eligible paths, npm gives the best install story for MCP clients. OCI would force users to volume-mount the project being scanned; MCPB would mean one bundle carrying every platform's binary, with narrower client support.

Targets **v0.21.1** (rebased onto main; the CLI released twice while this sat open).

## What

- **`npm/`** — a thin `bomly-mcp` package, zero dependencies.
  - `scripts/postinstall.js` downloads `bomly_<version>_<goos>_<goarch>.{tar.gz,zip}` from the matching release, verifies it against that release's `SHA256SUMS`, and unpacks the binary plus the archive's `LICENSE`, `NOTICE`, and `licenses/` into `vendor/`. A mismatch fails the install — there is no unverified fallback.
  - `bin/bomly-mcp.js` execs `bomly mcp serve` with stdio inherited, so stdout stays pure JSON-RPC, and forwards SIGINT/SIGTERM/SIGHUP so the child never orphans.
  - `scripts/prepack.js` copies the repo `LICENSE` and `NOTICE` into the package at pack time.
  - `BOMLY_MCP_VERSION` pins a version; `BOMLY_MCP_SKIP_DOWNLOAD=1` skips the download for lint/CI installs.
- **`server.json`** — registry metadata, schema `2025-12-11`, naming `io.github.bomly-dev/bomly-cli` and the `bomly-mcp` npm package.
- **`glama.json`** — display name, description, and keywords for Glama's indexer, so that listing reads correctly on day one.
- **`release.yml`** — `publish-npm` then `publish-mcp-registry`, split so either half can be re-run alone. npm auth is Trusted Publishing (OIDC); no token anywhere.
- **`ci.yml`** — a `npm wrapper` job running the new tests and asserting `npm/package.json`, `server.json`, and `mcpName` agree.
- **`auto-version.yml`** — bumps the version in `npm/package.json` and `server.json` alongside `cmd/bomly/main.go`.
- Docs: an `npx` path in `docs/MCP.md` and the README's AI-agents section.

## Ordering, and why it matters

`publish-npm` runs `needs: publish`. Releases stay drafts until the `publish` job finalizes them, and draft-release assets are not publicly downloadable — so publishing npm any earlier would ship a package whose postinstall 404s. `publish-mcp-registry` then waits for npm's CDN before calling `mcp-publisher`, because the registry verifies ownership by fetching the package.

## Review round 1 — what changed

| Finding | Fix |
| --- | --- |
| `NPM_TOKEN` left in `~/.npmrc`, readable by later steps including `mcp-publisher` | **Token removed entirely** — switched to npm Trusted Publishing (OIDC) after npm's token screen flagged long-lived automation tokens. Nothing left to scope or leak |
| Not retryable after a partial success — `npm publish` of an immutable version blocks the registry step behind it | Split into two jobs; the npm job treats an already-published exact version as success and continues |
| `mcp-publisher` streamed from `releases/latest` and executed unverified in a job with `id-token: write` | Pinned to v1.8.0, verified twice before extraction: pinned SHA-256, and `cosign verify-blob` against the Sigstore bundle asserting the registry's own tagged release workflow as signer |
| Archive `LICENSE`/`NOTICE`/`licenses/**` discarded; npm tarball carried no license either | Both fixed — `prepack` for the package, `copyNotices` for the vendored binary. Covered by tests |
| PowerShell path interpolation breaks on apostrophes and is an injection surface | Extraction now runs a `-File` script with paths as real argv |
| No automated coverage for the JS installer/launcher | 16 tests in `npm/test`, wired into CI |
| `SHA256SUMS` undefined at first use (CodeRabbit) | Defined in `docs/MCP.md` |
| `persist-credentials` not disabled (CodeRabbit) | `false` on both checkouts |
| Signal listener not removed before re-raising (CodeRabbit) | Fixed; see below |

The signal one was worth the dig. The bug was not a hang: with our listener still attached, re-raising re-entered our own handler, the event loop drained, and the wrapper exited **0** — telling a supervisor the server shut down cleanly when it had actually been killed. Verified by probe, and the test that covers it fails against the pre-fix code:

```
FIXED:   {"code":null,"signal":"SIGTERM"}
PRE-FIX: {"code":0,"signal":null}
```

## Before this can run — needs you

**No npm token.** `publish-npm` uses [npm Trusted Publishing](https://docs.npmjs.com/trusted-publishers) — npm exchanges the workflow's OIDC identity for publish rights, so there is no secret to store, scope, or rotate. `publish-mcp-registry` authenticates the same way. This replaced an earlier `NPM_TOKEN` design after npm's own token screen flagged it.

Trusted publishing is configured per package on npmjs.com, so it cannot create a package that does not exist yet. Setup is one-time (full version in `dev-docs/CI.md`):

1. Publish `bomly-mcp` once by hand, from a checkout at an already-published release tag:

   ```bash
   cd npm && npm publish --access public
   ```

   From your machine with `npm login` — not CI, so no automation token ever has to exist. The version must match a published GitHub release, since the postinstall downloads that release's archive.

2. Set the trusted publisher at `npmjs.com/package/bomly-mcp/access` → repo `bomly-dev/bomly-cli`, workflow `release.yml`. It is on the *package* access page, not the account packages page.

3. Turn the jobs on:

   ```bash
   gh variable set PUBLISH_MCP_REGISTRY -R bomly-dev/bomly-cli --body true
   ```

`bomly-mcp` is unclaimed on npm. The registry namespace `io.github.bomly-dev/*` comes from this repo's identity, no secret needed.

One thing to check: `glama.json` lists `bomly-guy` under `maintainers`, because that is the account with commits here. If you submit to Glama under a different handle, add it.

## Verified locally against v0.21.1

- `npm pack`, then a clean install of the tarball into a scratch project: binary downloaded, checksum verified, `bomly version` → `0.21.1`, and `vendor/` contains `bomly`, `LICENSE`, `NOTICE`, `licenses/`.
- Packed tarball contains `LICENSE`, `NOTICE`, `README.md`, `bin/`, `scripts/` — and no `vendor/`.
- Real MCP handshake through `./node_modules/.bin/bomly-mcp`: `initialize` → `{"name":"bomly","version":"0.21.1"}`, `tools/list` → all four tools, `bomly_scan` → a `schema_version: "mcp/1"` result. All three stdout lines were JSON-RPC; the banner stayed on stderr.
- `npm test` → 16/16.
- `server.json` validates against the registry's `2025-12-11` schema (ajv).
- The exact `cosign verify-blob` command from the workflow → `Verified OK`, and the pinned SHA-256 matches the published v1.8.0 artifact.
- `make generate` produces no drift.

Not yet exercised: the publish jobs themselves, which cannot run until the secret and variable above exist.


bomly-guy (commented):
Found five actionable issues in the release/npm path; details are inline. The new JavaScript installer and launcher also have no automated coverage yet—please add focused tests for checksum parsing/failure, archive contents, platform mapping, npm package contents, and child-process shutdown behavior.

.github/workflows/release.yml
```diff
@@ -226,3 +226,88 @@ jobs:
...
+      - name: Publish bomly-mcp to npm
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          set -euo pipefail
+          if [ -z "${NPM_TOKEN}" ]; then
+            echo "::error::PUBLISH_MCP_REGISTRY is true but NPM_TOKEN is not set. Run: gh secret set NPM_TOKEN -R bomly-dev/bomly-cli"
+            exit 1
+          fi
+          echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
```
bomly-guy: Please avoid leaving the publish credential in `~/.npmrc`. Step-scoped environment variables do not scope files, so every later step—including the downloaded `mcp-publisher` binary—can read this token. Use a temporary `NPM_CONFIG_USERCONFIG` (or a setup-node token placeholder) and remove it immediately after `npm publish`, ideally with a trap.

.github/workflows/release.yml
```diff
@@ -226,3 +226,88 @@ jobs:
...
+          npm publish --provenance --access public ./npm
```
bomly-guy: This makes the registry publication impossible to retry after a partial success. If `npm publish` succeeds but the CDN wait, OIDC login, or registry publish fails, rerunning the failed job attempts to publish the same immutable npm version and stops here before reaching the registry. Please make this idempotent: detect an already-published exact version, verify it is the expected artifact, and continue, or split npm and registry publication into independently rerunnable jobs.

.github/workflows/release.yml
```diff
@@ -226,3 +226,88 @@ jobs:
...
+            echo "Attempt ${attempt}: bomly-mcp@${version} not visible yet, waiting."
+            sleep 10
+          done
+          echo "::error::bomly-mcp@${version} did not become resolvable on npm within 5 minutes"
+          exit 1
+
+      - name: Install mcp-publisher
+        run: |
+          set -euo pipefail
+          curl -fsSL "https://github.com/modelcontextprotocol/registry/releases/latest/download/mcp-publisher_linux_amd64.tar.gz" \
```
bomly-guy: This downloads and immediately executes a moving `latest` asset without verifying its digest or signature, while the job has `id-token: write`. An upstream release change or compromise would therefore run arbitrary code in the privileged release job. Please pin an explicit `mcp-publisher` version and verify the release-provided SHA-256/Sigstore material before extracting and running it.

npm/scripts/postinstall.js
```diff
@@ -0,0 +1,138 @@
...
+    unpack(archivePath, staging, ext);
+
+    const extracted = path.join(staging, binary);
+    if (!fs.existsSync(extracted)) {
+      fail(`${archiveName} did not contain ${binary}`);
+    }
+
+    fs.mkdirSync(VENDOR_DIR, { recursive: true });
+    const installed = path.join(VENDOR_DIR, binary);
+    fs.copyFileSync(extracted, installed);
```
bomly-guy, line 167: The GoReleaser archive includes `LICENSE`, `NOTICE`, and `licenses/**`, but this installer discards all of them and copies only the executable. Separately, `npm pack --dry-run --json` on this commit shows that the npm tarball itself has no license/notice files. That leaves the installed package redistributing Bomly and its bundled components without the accompanying notices. Please include the repository license/notice in the npm package and preserve the release archive's applicable notice/license files alongside the vendored binary.

npm/scripts/postinstall.js
```diff
@@ -0,0 +1,138 @@
...
+  if (ext === "zip") {
+    // PowerShell ships with every supported Windows version; `tar` does too on
+    // Windows 10 1803+, but Expand-Archive is the safer floor for zip.
+    execFileSync(
+      "powershell",
+      [
+        "-NoProfile",
+        "-NonInteractive",
+        "-Command",
+        `Expand-Archive -LiteralPath '${archivePath}' -DestinationPath '${destination}' -Force`,
```
bomly-guy: Interpolating these paths into a single-quoted PowerShell command breaks for valid Windows temp/profile paths containing an apostrophe (for example `C:\Users\O'Brien\...`) and also creates an avoidable command-injection surface. Please pass the paths as separate PowerShell arguments (referenced through `$args`) or escape them as PowerShell literals instead of embedding them in command text.

bomly-guy:
All six addressed in 9619fb0, plus the version bump to 0.21.1.

**Credential scoping.** You were right that `env:` does not scope a file. The token now goes to `$RUNNER_TEMP/.npmrc` selected by `NPM_CONFIG_USERCONFIG`, written under `umask 077`, removed by a `trap ... EXIT` so a failed publish still cleans up. Beyond that I split the job in two: `publish-mcp-registry` is now a separate job, so the step that runs the downloaded `mcp-publisher` binary never receives `NPM_TOKEN` in its environment at all.

**Retryability.** Split into `publish-npm` → `publish-mcp-registry`, and the npm job now checks the registry for the exact version first and exits 0 if it is already live. So a re-run after a partial success walks through the npm half and reaches the registry step instead of dying on the immutable version.

**Pinning and verification.** Pinned to `v1.8.0` with two independent checks before anything is extracted or executed: the pinned SHA-256, and `cosign verify-blob` against the release's Sigstore bundle with `--certificate-identity` asserting `modelcontextprotocol/registry/.github/workflows/release.yml@refs/tags/v1.8.0` and `--certificate-oidc-issuer` the GitHub Actions issuer. I ran that exact command locally against the published artifact — `Verified OK`. Version, digest, and identity are three `env:` values at the top of the job so a bump is one reviewed change.

**Licenses and notices.** Both halves fixed. `scripts/prepack.js` copies the repo `LICENSE` and `NOTICE` into the package at pack time (gitignored inside `npm/`, so there is no second copy to drift), and `copyNotices` preserves the archive's `LICENSE`, `NOTICE`, and `licenses/` next to the vendored binary. Confirmed on a clean install: `vendor/` now holds `bomly`, `LICENSE`, `NOTICE`, `licenses/`, and `npm pack` lists `LICENSE` and `NOTICE`. Both are covered by tests.

**PowerShell quoting.** Replaced with a `-File` script taking `-Archive` and `-Destination` as parameters, so the paths arrive as real argv entries and never touch a command string. The apostrophe case and the injection surface both go away.

**Tests.** Added 16 under `npm/test`, run by a new `npm wrapper` CI job, covering everything you listed: platform mapping including unsupported targets, `SHA256SUMS` parsing (prefix-colliding names, missing entry, empty file), checksum sensitivity, archive extraction, notice propagation, launcher argument pass-through, a clean stdout, exit-code propagation, both signal paths, and the packed tarball's contents. CI also asserts `npm/package.json`, `server.json`, and `mcpName` agree, which is the drift most likely to slip through.

On CodeRabbit's signal comment — worth flagging, because the diagnosis in the thread is not quite what was happening. It is not a hang. With our listener still attached, re-raising re-entered our own handler, the event loop then drained, and the wrapper exited **0**: a supervisor was told the server shut down cleanly when it had in fact been killed. Probed both ways:

```
FIXED:   {"code":null,"signal":"SIGTERM"}
PRE-FIX: {"code":0,"signal":null}
```

The test asserting that fails against the pre-fix code, so it is guarding the real behavior rather than just passing.

PR #341, opened 2026-07-27
Bound in-process repository file reads

bomly-guy:
## What changed

- apply a 64 MiB limit before in-process parsers read repository manifests, lockfiles, workspace metadata, analyzer inputs, or source files
- enforce the same limit while streaming so a file that grows after it is opened cannot bypass the check
- treat corrupt or oversized matcher and analyzer cache entries as cache misses
- add a source-level assurance test that prevents unbounded whole-file reads from returning to detector, analyzer, and registry code
- document the covered readers, existing specialized limits, and intentional command-output exclusions

## Why

Repository files are untrusted input. Previously, many parsers loaded a complete file with no byte limit. A hostile or accidental oversized file could consume substantial memory before validation began.

The shared ceiling is deliberately generous for large monorepo lockfiles. Bomly rejects an over-limit document instead of parsing a prefix, so it cannot report a believable but incomplete dependency graph. Package-manager command output remains untruncated for the same completeness reason.

## Validation

- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `make fuzz FUZZTIME=5s`
- `go test ./internal/system -run '^$' -bench '^BenchmarkReadLimit$' -benchtime=1x`
- focused detector, analyzer, cache, system, registry, and assurance suites

## Summary by CodeRabbit

* **New Features**
  * Added a shared 64 MiB per-file limit for repository manifests, lockfiles, workspace metadata, and analyzer source files.
  * Added equivalent limits for local cache entries; oversized or corrupted entries now safely behave as cache misses.
  * Large inputs are rejected before parsing, including files that exceed limits during streaming reads.

* **Documentation**
  * Expanded security and assurance documentation covering input limits, discovery constraints, cache handling, and residual resource risks.
  * Added repository input-limit guidance and supporting regression coverage.


bomly-guy:
Reviewed. This is a well-executed sweep, and the design choices that matter are right:

- `OpenFileLimit` checks the stat size up front *and* enforces the budget while streaming, with a 1-byte probe distinguishing exactly-at-limit from over — same discipline as the plugin-archive work in #332, now pinned by `TestOpenFileLimitRejectsFileGrowthWhileStreaming`, which is the case a stat-only check silently loses.
- The reject-don't-truncate policy ("parsers do not use a partial document or partial graph") is the correct call for a security tool and is stated in the assurance doc rather than left implicit. Same for the deliberate non-truncation of package-manager output.
- Oversized/corrupt cache entries degrade to misses instead of failing the scan, with `TestFileCache_OversizedEntryIsGracefulMiss` covering it — consistent with the repo's cache-failures-are-non-fatal rule.
- `*os.PathError` passthrough on open is preserved, so the `errors.Is(err, os.ErrNotExist)` contract from the #336 follow-up keeps working at every converted call site (I checked `moduleChecksum` and the discovery probes — the not-exist branches still fire).
- The docs are honest about the layering: the exclusions table (yarn head read, command output, directory walking) says what is *not* covered instead of implying blanket coverage.

I also verified coverage on the branch: zero remaining `os.ReadFile`/`io.ReadAll` calls in `internal/analyzers`, `internal/detectors`, `internal/registry` outside tests; the one bare `os.Open` (yarn format head-read) is bounded to a fixed head and documented as an intentional exclusion.

**One gap worth closing: the assurance tripwire only greps `os.ReadFile(`.** The exact evasion it should catch — `os.Open` + `io.ReadAll` — is invisible to it, and the yarn head-reader shows bare `os.Open` already lives legitimately in the guarded tree, so the pattern will recur. Since the guarded directories currently have zero `io.ReadAll(` hits, extending the pattern list to `io.ReadAll(` (and `ioutil.ReadFile(` for completeness) costs nothing today and turns the test from "we renamed the function" into "we bounded the read". A `// assurance:allow-unbounded-read <reason>` escape-comment convention would keep future legitimate exceptions explicit rather than silently narrowing the check.

Two trivia, no action needed: `TestRepositoryAndCacheFileLimits` asserts a constant equals its own literal — fine as a change-detector, but it can only fail when someone edits both sides inconsistently; and `limitedFileReader` re-probes one byte on every post-limit `Read` call, which is harmless given `io.ReadAll` stops at the first error.

Build + focused suites green on the branch head locally. Nothing blocking from me once the tripwire pattern is widened.

PR #342, opened 2026-07-27
Bound remote Git target materialization

bomly-guy:
## What changed

- give remote Git clone and checkout work a 10-minute deadline
- stop implicit submodule and Git LFS object downloads
- remove partial temporary clones after cancellation or failure
- validate a completed checkout before discovery with limits of 1,000,000 paths, 10 GiB of regular files, and 256 path levels
- allow repository symlinks that stay inside the checkout and reject links that escape it
- document which Git transfer and object-storage behavior still requires an operating-system quota

## Why

`--url` deliberately gives Git network and disk authority, but Bomly previously placed no time or checkout boundary around it. A remote repository could also commit a manifest symlink that caused an in-process detector to read outside the temporary checkout.

The new checks cover the controls Bomly can enforce portably. Git transfer bytes and `.git` object storage can be consumed before checkout validation, so the documentation does not claim a byte quota that Git cannot guarantee.

## Dependency

This draft is stacked on #341 so the final security documentation describes both input and remote-target limits without a merge conflict. Review this PR by comparing it with `harden-repository-input-budgets`.

## Validation

- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- focused Git, CLI options, and CLI suites before and after stacking
- cancellation cleanup, exact checkout limits, contained links, escaping links, and secret-safe error regressions


bomly-guy:
Reviewed against the `harden-repository-input-budgets` base as the description asks. The core mechanics are sound: the cancellation-cleanup test, the containment errors that name the link but deliberately never echo the outside target (and assert that), and honest docs that Git can consume transfer bytes and `.git` storage *before* the checkout validation runs — no fake byte-quota claim. The symlink policy is also more precise than a blanket ban: contained links keep working, and the lexical resolution is conservative in the right direction (a `link -> dir-link/../../x` collapses lexically and gets rejected; a chain that escapes through another in-tree link is caught because every link in the tree is individually validated). I checked those cases against `validateMaterializedTree` and the escape logic holds.

Two substantive findings:

**1. The LFS/submodule suppression does not survive past the clone invocation.**

```go
args = append(args, "-c", "filter.lfs.smudge=", "-c", "filter.lfs.required=false")
...
args = append(args, "clone", "--quiet", "--no-recurse-submodules")
```

The `-c` pairs are placed *before* the subcommand, i.e. `git -c … clone …` — per-invocation config only. (After the subcommand, `git clone -c/--config` would persist into the new repo's config; before it, it doesn't.) So the clone's own HEAD checkout is protected, but when `--ref` is set, `cloneInto` follows with `checkoutCommitContext` → `git checkout --quiet --detach <commit>` with **no** suppression flags. On a machine with `git lfs install` in global config, a repository whose ref tree contains LFS-tracked files resumes exactly the unbounded download this PR exists to stop. Same shape for `submodule.recurse=true` in user config: `--no-recurse-submodules` rides only the clone, and `git checkout` honors that config too. The `bomly diff --url` path has a third hop with the same hole — `MaterializeLocalRef` does a fresh `clone --local`, which does not copy the source clone's config, so even the persist-into-repo variant wouldn't cover it.

The robust fix is environmental rather than positional: set `GIT_LFS_SKIP_SMUDGE=1` (the canonical git-lfs kill-switch) plus `-c submodule.recurse=false` on every materialization invocation — clone *and* checkout — at the `runGitContext` level or via a dedicated arg/env builder those call sites share. And a regression test that asserts the checkout invocation carries the suppression, since this is precisely the sort of thing a refactor reorders back.

**2. Remote containment is being applied to trusted local repositories, and a pinned behavior was reversed silently.**

`validateMaterializedTree` runs in `MaterializeLocalRef`, which is the path for plain local `bomly diff --base/--head`. That means the user's own repository history is now subject to the escape-symlink rejection and the 1M-path/10 GiB/256-depth limits. Two problems with that:

- It contradicts the trust model this series has been carefully building — #330's explicit-vs-automatic distinction, `SECURITY.md`'s "local target size remains under the user's control" — and produces a visible inconsistency: a repo with a committed absolute symlink (dotfiles/Nix/home-manager layouts, monorepos linking `/opt` SDKs) scans fine via `--path` but now fails `bomly diff HEAD~1` with a containment error.
- #326 merged `TestMaterializeLocalRefKeepsRepositorySymlinksAsSymlinks` pinning exactly this: a symlink to a target *outside* the repo survives materialization. This PR rewrites that test to an in-repo relative link — the pinned behavior is reversed without being mentioned in the PR description or docs, which frame everything as `--url`.

The wrinkle is that `MaterializeLocalRef` *does* need validation on one path: `bomly diff --url` clones remotely once, then materializes each ref through this function, so an escaping symlink present only at `--base` would slip past a CloneTemp-only check. The fix is to thread origin trust: `resolveDiffRepo` already knows whether `repoRoot` came from `CloneTemp` (`current.URL != ""`), so pass validate/skip (or validate/warn) down accordingly. If you instead decide local diff *should* be contained, that's defensible — but then it's a breaking behavior change that needs to be stated in the PR body, `docs/SCAN_TARGETS.md` (which currently scopes all of this to "Scanning a remote repository"), and the #326 test change called out as deliberate.

**Minor:**

- `RemoteOperationTimeout`'s comment says it "bounds each Git clone or checkout operation", but one 10-minute deadline spans clone + resolve + checkout inside `cloneTemp`, and `MaterializeLocalRef` has no deadline at all (fine for disk-local ops — just make the comment match reality).
- The hidden benchmark still shells out its own `git clone --depth 1` (`internal/benchmark/run.go`) outside all of these boundaries — same second-construction-site class as the benchmark stderr gate on #334. Worth a line in the residual-risk docs or a follow-up, not a blocker for a hidden command.
- The validation adds a full tree walk per materialization ahead of discovery's own walk; at the 1M-entry ceiling that's real but acceptable cost for the boundary.

Build + git/opts/assurance suites green on the branch head locally; CI green on both PRs. Finding 1 I'd consider blocking for the PR's own threat model; finding 2 needs either the trust-threading or an explicit, documented decision before this leaves draft.

PR #347, opened 2026-07-28
Report dependency detail changes in diff

bomly-guy:
## Summary

- classify dependency relationship, source, and registry-matching eligibility changes separately from package version changes
- preserve lean before-and-after occurrence evidence in diff JSON and generated schemas
- show the same dependency detail changes in text, Markdown, the interactive diff view, and compact MCP responses
- preserve duplicate manifest occurrences, fuzzy-match evidence, and head-side package enrichment

## Why

A dependency can keep the same identity and version while moving from direct to transitive, changing from a registry source to Git or a workspace, or becoming ineligible for registry matching. These changes were previously invisible or could be mistaken for package additions and removals.

The canonical classifier now lives in the SDK comparison layer. Version changes and dependency detail changes remain parallel results, so existing version-change meaning stays intact and every output surface projects the same evidence.

## User impact

Diff reports now include a dedicated dependency detail-change count and before/after evidence. Human-readable formats use plain labels such as “Dependency detail changes,” while JSON and MCP retain stable structured transition records. CLI schema `1.0`, MCP schema `mcp/1`, and plugin protocol `v1` remain unchanged.

## Validation

- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`

## Summary by CodeRabbit

* **New Features**
  * Diff reports now distinguish package version changes from dependency detail changes.
  * Dependency detail changes include relationship, source, and registry-matching eligibility transitions with before-and-after information.
  * Text, Markdown, JSON, interactive, and MCP outputs now display dependency transitions consistently.
  * Interactive views show transitioned dependencies with dedicated counts, filters, and details.
* **Documentation**
  * Updated architecture, model, output-format, and pipeline documentation to describe the expanded diff results.

### DIFF
--- a/internal/cli/render/diff.go
+++ b/internal/cli/render/diff.go
@@ -55,7 +55,7 @@ func dependencyTextSections(results output.DiffDependencyResults) []string {
 		if len(results.Changed) == 0 {
 			return
 		}
-		lines = append(lines, Style(fmt.Sprintf("Changed (%d)", len(results.Changed)), Bold))
+		lines = append(lines, Style(fmt.Sprintf("Version changed (%d)", len(results.Changed)), Bold))
 		for _, change := range results.Changed {
 			name := change.After.Name
 			if strings.TrimSpace(name) == "" {
@@ -65,13 +65,66 @@ func dependencyTextSections(results output.DiffDependencyResults) []string {
 			lines = append(lines, Wrap(line, Yellow))
 		}
 	}
+	appendTransitions := func() {
+		if len(results.Transitions) == 0 {
+			return
+		}
+		lines = append(lines, Style(fmt.Sprintf("Dependency detail changes (%d)", len(results.Transitions)), Bold))
+		for _, transition := range results.Transitions {
+			name := dependencyTransitionDisplayName(transition.After)
+			line := fmt.Sprintf("  ↔ %s  %s", name, dependencyTransitionDescription(transition))
+			lines = append(lines, Wrap(line, Yellow))
+		}
+	}
 	appendAdded()
 	appendRemoved()
 	appendChanged()
+	appendTransitions()
 	if len(lines) == 0 {
 		lines = append(lines, Style("No dependency changes.", Dim))
 	}
 	return lines
+}
+
+func dependencyTransitionDescription(transition output.DiffDependencyTransition) string {
+	parts := make([]string, 0, len(transition.ChangedFields))
+	for _, field := range transition.ChangedFields {
+		switch field {
+		case sdk.DependencyMetadataRelationship:
+			parts = append(parts, fmt.Sprintf(
+				"relationship: %s → %s",
+				valueOrDash(transition.Before.Relationship),
+				valueOrDash(transition.After.Relationship),
+			))
+		case sdk.DependencyMetadataSource:
+			parts = append(parts, fmt.Sprintf(
+				"source: %s → %s",
+				valueOrDash(string(transition.Before.Source)),
+				valueOrDash(string(transition.After.Source)),
+			))
+		case sdk.DependencyMetadataRegistryEligibility:
+			parts = append(parts, fmt.Sprintf(
+				"registry matching: %s → %s",
+				registryEligibilityLabel(transition.Before.RegistryEligible),
+				registryEligibilityLabel(transition.After.RegistryEligible),
+			))
+		}
+	}
+	return strings.Join(parts, "; ")
+}
+
+func dependencyTransitionDisplayName(state output.DiffDependencyTransitionState) string {
+	if name := strings.TrimSpace(state.Name); name != "" {
+		return name
+	}
+	return state.ID
+}
+
+func registryEligibilityLabel(eligible bool) string {
+	if eligible {
+		return "eligible"
+	}
+	return "not eligible"
 }
 
 // findingsSummaryLine produces a summary line plus a list of each introduced or
--- a/internal/cli/render/diff.go
+++ b/internal/cli/render/diff.go
@@ -69,7 +69,7 @@ func dependencyTextSections(results output.DiffDependencyResults) []string {
 		if len(results.Transitions) == 0 {
 			return
 		}
-		lines = append(lines, Style(fmt.Sprintf("Dependency detail changes (%d)", len(results.Transitions)), Bold))
+		lines = append(lines, Style(fmt.Sprintf("Detail changes (%d)", len(results.Transitions)), Bold))
 		for _, transition := range results.Transitions {
 			name := dependencyTransitionDisplayName(transition.After)
 			line := fmt.Sprintf("  ↔ %s  %s", name, dependencyTransitionDescription(transition))
@@ -88,29 +88,42 @@ func dependencyTextSections(results output.DiffDependencyResults) []string {
 
 func dependencyTransitionDescription(transition output.DiffDependencyTransition) string {
 	parts := make([]string, 0, len(transition.ChangedFields))
+	sourceChanged := dependencyDetailFieldChanged(transition, sdk.DependencyDetailSource)
 	for _, field := range transition.ChangedFields {
 		switch field {
-		case sdk.DependencyMetadataRelationship:
+		case sdk.DependencyDetailRelationship:
 			parts = append(parts, fmt.Sprintf(
 				"relationship: %s → %s",
-				valueOrDash(transition.Before.Relationship),
-				valueOrDash(transition.After.Relationship),
+				valueOrDash(string(transition.Before.Relationship)),
+				valueOrDash(string(transition.After.Relationship)),
 			))
-		case sdk.DependencyMetadataSource:
+		case sdk.DependencyDetailSource:
 			parts = append(parts, fmt.Sprintf(
 				"source: %s → %s",
 				valueOrDash(string(transition.Before.Source)),
 				valueOrDash(string(transition.After.Source)),
 			))
-		case sdk.DependencyMetadataRegistryEligibility:
+		case sdk.DependencyDetailRegistryEligibility:
+			if sourceChanged {
+				continue
+			}
 			parts = append(parts, fmt.Sprintf(
-				"registry matching: %s → %s",
+				"vulnerability checks: %s → %s",
 				registryEligibilityLabel(transition.Before.RegistryEligible),
 				registryEligibilityLabel(transition.After.RegistryEligible),
 			))
 		}
 	}
 	return strings.Join(parts, "; ")
+}
+
+func dependencyDetailFieldChanged(transition output.DiffDependencyTransition, wanted sdk.DependencyDetailField) bool {
+	for _, field := range transition.ChangedFields {
+		if field == wanted {
+			return true
+		}
+	}
+	return false
 }
 
 func dependencyTransitionDisplayName(state output.DiffDependencyTransitionState) string {
@@ -122,9 +135,9 @@ func dependencyTransitionDisplayName(state output.DiffDependencyTransitionState)
 
 func registryEligibilityLabel(eligible bool) string {
 	if eligible {
-		return "eligible"
+		return "covered"
 	}
-	return "not eligible"
+	return "not covered"
 }
 
 // findingsSummaryLine produces a summary line plus a list of each introduced or
--- a/internal/cli/render/diff_markdown.go
+++ b/internal/cli/render/diff_markdown.go
@@ -56,7 +56,13 @@ func diffOverviewMarkdown(payload output.DiffResponse) []string {
 		[][]string{{
 			status,
 			fmt.Sprintf("+%d / ~%d / -%d", payload.Summary.AddedManifestCount, payload.Summary.ChangedManifestCount, payload.Summary.RemovedManifestCount),
-			fmt.Sprintf("+%d / ~%d / -%d", payload.Summary.AddedPackageCount, payload.Summary.ChangedPackageCount, payload.Summary.RemovedPackageCount),
+			fmt.Sprintf(
+				"+%d / ~%d version / ~%d details / -%d",
+				payload.Summary.AddedPackageCount,
+				payload.Summary.ChangedPackageCount,
+				payload.Summary.TransitionedPackageCount,
+				payload.Summary.RemovedPackageCount,
+			),
 			fmt.Sprintf("%d introduced / %d persisted / %d resolved", introduced, persisted, resolved),
 			humanizeDurationMS(payload.Metadata.DurationMS),
 		}},
@@ -79,16 +85,47 @@ func humanizeDurationMS(ms int64) string {
 func diffDependencyMarkdown(payload output.DiffResponse) []string {
 	results := payload.Results.Dependencies
 	lines := []string{
-		fmt.Sprintf("**Summary:** %d added, %d changed, %d removed.", len(results.Added), len(results.Changed), len(results.Removed)),
+		fmt.Sprintf(
+			"**Summary:** %d added, %d version changed, %d with detail changes, %d removed.",
+			len(results.Added), len(results.Changed), len(results.Transitions), len(results.Removed),
+		),
 		"",
 	}
 	lines = append(lines, diffAddedRemovedDependencyTable("Added Dependencies", "added", results.Added)...)
 	lines = append(lines, diffChangedDependencyTable(results.Changed)...)
+	lines = append(lines, diffDependencyTransitionTable(results.Transitions)...)
 	lines = append(lines, diffAddedRemovedDependencyTable("Removed Dependencies", "removed", results.Removed)...)
-	if len(results.Added) == 0 && len(results.Changed) == 0 && len(results.Removed) == 0 {
+	if len(results.Added) == 0 && len(results.Changed) == 0 && len(results.Transitions) == 0 && len(results.Removed) == 0 {
 		return []string{"✅ No dependency changes."}
 	}
 	return trimTrailingMarkdownBlanks(lines)
+}
+
+func diffDependencyTransitionTable(transitions []output.DiffDependencyTransition) []string {
+	if len(transitions) == 0 {
+		return nil
+	}
+	sorted := append([]output.DiffDependencyTransition(nil), transitions...)
+	sort.Slice(sorted, func(i, j int) bool {
+		left := sorted[i].After
+		right := sorted[j].After
+		if dependencyTransitionDisplayName(left) != dependencyTransitionDisplayName(right) {
+			return dependencyTransitionDisplayName(left) < dependencyTransitionDisplayName(right)
+		}
+		return left.ID < right.ID
+	})
+	rows := make([][]string, 0, len(sorted))
+	for _, transition := range sorted {
+		rows = append(rows, []string{
+			dependencyTransitionDisplayName(transition.After),
+			valueOrDash(transition.After.Version),
+			dependencyTransitionDescription(transition),
+		})
+	}
+	return append(
+		[]string{"### Dependency Detail Changes", ""},
+		append(markdownTable([]string{"Package", "Version", "Changes"}, rows), "")...,
+	)
 }
 
 func diffAddedRemovedDependencyTable(title, status string, changes []output.DiffPackageChange) []string {
--- a/internal/cli/render/diff_markdown.go
+++ b/internal/cli/render/diff_markdown.go
@@ -57,7 +57,7 @@ func diffOverviewMarkdown(payload output.DiffResponse) []string {
 			status,
 			fmt.Sprintf("+%d / ~%d / -%d", payload.Summary.AddedManifestCount, payload.Summary.ChangedManifestCount, payload.Summary.RemovedManifestCount),
 			fmt.Sprintf(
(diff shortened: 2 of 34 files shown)


internal/output/view.go
```diff
@@ -134,31 +135,54 @@ type DiffChangedPackage struct {
...
-	ExactMatchCount        int `json:"exact_match_count"`
-	FuzzyMatchCount        int `json:"fuzzy_match_count"`
-	UnmatchedPackageCount  int `json:"unmatched_package_count"`
+	AddedManifestCount       int `json:"added_manifest_count"`
+	ChangedManifestCount     int `json:"changed_manifest_count"`
+	RemovedManifestCount     int `json:"removed_manifest_count"`
+	UnchangedManifestCount   int `json:"unchanged_manifest_count"`
+	AddedPackageCount        int `json:"added_package_count"`
+	ChangedPackageCount      int `json:"changed_package_count"`
+	TransitionedPackageCount int `json:"transitioned_package_count"`
```
bomly-guy, line 181: **Blocking: smoke goldens will drift.** This field has no `omitempty`, so `transitioned_package_count` now serializes in every diff JSON document. Six existing goldens contain the summary block (`diff-npm`, `diff-go`, `diff-go-audit`, `diff-sbom`, `lite-diff-go`, `container-diff-alpine`) and none are regenerated in this PR — the nightly smoke run will fail on all of them.

Please run `make smoke ARGS="-update"` (or dispatch **Update Smoke Goldens**) and commit the regenerated goldens in this PR. Per the feature checklist, a user-visible feature should also get a smoke case that exercises it meaningfully — none of the current diff cases actually produces a transition, so the end-to-end path (real repo → transition in JSON/text) is untested.

sdk/graph.go
```diff
@@ -428,9 +468,100 @@ func Compare(base, head *Graph) Diff {
...
 	})
+	sortDependencyMetadataTransitions(diff.Transitions)
 	return diff
 }
 
+// CompareDependencyMetadata returns a transition when relationship, source, or
+// registry-matching eligibility differs between two occurrences. It is
+// exported so trusted fuzzy identity reconciliation can use the same canonical
+// classifier as Compare.
+func CompareDependencyMetadata(baseGraph, headGraph *Graph, before, after *Dependency) (DependencyMetadataTransition, bool) {
```
bomly-guy: **Blocking for real-world usability: cross-detector diffs will flood the report.** This compares `Source` string-for-string and infers relationship from graph shape when unset. The SBOM detector never sets `Source`, while the node lockfile detectors set `registry`/`git`/etc. explicitly. So diffing an SBOM file against a source scan of the same tree — a supported, advertised workflow (cf. the `diff-sbom` smoke case) — emits a `source: – → registry` and/or `relationship: unknown → direct` transition for nearly every matched package. The one transition the user cares about drowns in hundreds of artifacts of detector asymmetry.

From a user's perspective, a diff that says "487 dependency detail changes" when they changed one lockfile line destroys trust in the signal on first contact. Suggest suppressing a field from `changedFields` when **either side's value is empty/unknown** (absence of evidence isn't evidence of change):

```go
if before.Source != "" && after.Source != "" && before.Source != after.Source {
	changedFields = append(changedFields, DependencyMetadataSource)
}
```

and analogously only report a relationship change when neither side resolved to `DependencyRelationshipUnknown`. That keeps same-detector diffs (the overwhelmingly common case) fully covered while making cross-format diffs quiet by default.

sdk/graph.go
```diff
@@ -428,9 +468,100 @@ func Compare(base, head *Graph) Diff {
...
+		After:                  after,
+		ChangedFields:          changedFields,
+		BeforeRelationship:     beforeRelationship,
+		AfterRelationship:      afterRelationship,
+		BeforeRegistryEligible: beforeEligible,
+		AfterRegistryEligible:  afterEligible,
+	}, true
+}
+
+func dependencyRelationshipForDiff(graph *Graph, node *Dependency) DependencyRelationship {
```
bomly-guy: Minor perf: this runs per matched pair, and on legacy graphs without explicit relationships each call re-runs `Roots()` plus `DirectDependencies(root)` — which re-sorts the child list every time. For a large npm graph that's O(N · roots · D log D) per manifest side. Worth hoisting a per-graph `map[nodeID]DependencyRelationship` computed once (roots' direct children → direct, other nodes with parents → transitive) and passing it in, since `Compare` already holds both graphs.

sdk/graph.go
```diff
@@ -37,6 +38,34 @@ type VersionChange struct {
...
+	// project source change.
+	DependencyMetadataSource DependencyMetadataField = "source"
+	// DependencyMetadataRegistryEligibility indicates that external registry
+	// matching eligibility changed.
+	DependencyMetadataRegistryEligibility DependencyMetadataField = "registry_eligibility"
+)
+
+// DependencyMetadataTransition captures same-identity occurrence metadata
+// changes. Version changes remain represented separately by VersionChange.
+type DependencyMetadataTransition struct {
```
bomly-guy: Naming nit: `DependencyMetadataTransition` collides with `Dependency.Metadata` — the actual map field — which this classifier does **not** compare. Someone reading the SDK will reasonably assume this diffs `dep.Metadata`. Since the human-facing surfaces already settled on "detail changes", consider `DependencyDetailTransition` (or `OccurrenceTransition`), which also tightens the SDK↔UI vocabulary.

internal/cli/render/diff.go
```diff
@@ -65,15 +65,68 @@ func dependencyTextSections(results output.DiffDependencyResults) []string {
...
 	appendRemoved()
 	appendChanged()
+	appendTransitions()
 	if len(lines) == 0 {
 		lines = append(lines, Style("No dependency changes.", Dim))
 	}
 	return lines
 }
 
+func dependencyTransitionDescription(transition output.DiffDependencyTransition) string {
```
bomly-guy, line 89: **Main UX issue with the concept:** `RegistryMatchEligible()` is derived almost entirely from `Source`, so the common case (registry → git) renders as:

> `source: registry → git; registry matching: eligible → not eligible`

— the same fact twice, the second phrased in internal jargon ("registry matching" means nothing to a user who hasn't read the engine code). The test in this PR asserts exactly this triple-report for one conceptual change.

Two suggestions, both purely presentational (keep JSON/MCP structured fields as-is):

1. Skip the eligibility line when `ChangedFields` also contains `source` — it's implied.
2. When eligibility *did* change independently, word it as the consequence the user should act on, e.g.:

```go
func registryEligibilityLabel(eligible bool) string {
	if eligible {
		return "covered"
	}
	return "no longer covered"
}
```

with the label `vulnerability checks:` instead of `registry matching:`. "vulnerability checks: covered → no longer covered" tells the user *why they should care* — Bomly can no longer match this package against vulnerability databases — which is the strongest part of this whole feature and currently the most obscured.

internal/cli/render/diff.go
```diff
@@ -65,15 +65,68 @@ func dependencyTextSections(results output.DiffDependencyResults) []string {
 			lines = append(lines, Wrap(line, Yellow))
 		}
 	}
+	appendTransitions := func() {
+		if len(results.Transitions) == 0 {
+			return
+		}
+		lines = append(lines, Style(fmt.Sprintf("Dependency detail changes (%d)", len(results.Transitions)), Bold))
```
bomly-guy: "Dependency detail changes" is a reasonable human label — but it's currently one of **six** names for the same concept: `transitions` (JSON), `dependency_transitions` + `packages_transitioned` (MCP), "Details changed" (TUI counts), `TRANSITIONED` (TUI badge), "metadata transitions" (docs/SDK). Suggest standardizing: **"Detail changes"** as the one human-facing term everywhere (it pairs naturally with "Version changed" and this section header already leads with it), `transitions` as the one stable JSON noun, and retiring "transitioned" from anything a user sees. Whatever term wins, the docs page should introduce it once and use it consistently — right now `docs/OUTPUT_FORMATS.md` introduces the concept as "metadata transition" while this renderer calls it "detail changes", so a user cross-referencing docs and output can't tell they're the same feature.

internal/tui/utils.go
```diff
@@ -158,6 +158,8 @@ func statusBadge(status string) string {
 		return terminalSafeBadge(label, render.BgRed, render.White)
 	case "changed":
 		return terminalSafeBadge(label, render.BgYellow, render.Black)
+	case "transitioned":
```
bomly-guy, line 161: The `TRANSITIONED` badge is the weakest wording in the PR: standing alone in a row, it tells a user nothing (transitioned to *what?* from *what?*). Since the badge text derives from the status string, consider mapping this status to a plainer display label — `DETAILS` or `CHANGED*` — or keep the internal status `transitioned` but render the badge as `DETAILS CHANGED` truncated to badge width. Also worth a color distinct from `unchanged`: right now both render on `BgNeutral`, so the *only* new change-kind and "nothing happened" look identical at a glance.

internal/tui/diff.go
```diff
@@ -382,7 +382,7 @@ func (m *DiffModel) diffAggregateCounts() diffAggregateCounts {
 	s := m.payload.Summary
 	out := diffAggregateCounts{
 		ManifestDeltas: s.AddedManifestCount + s.ChangedManifestCount + s.RemovedManifestCount,
-		PackageDeltas:  s.AddedPackageCount + s.ChangedPackageCount + s.RemovedPackageCount,
+		PackageDeltas:  s.AddedPackageCount + s.ChangedPackageCount + s.TransitionedPackageCount + s.RemovedPackageCount,
```
bomly-guy: **Double counting in every aggregate view.** A package whose version *and* relationship both changed is in `Changed` **and** `Transitions`, and this sum (plus `changedTotal`, the ecosystem histogram, and the per-manifest "top changed" ranking) now counts it twice. In the components tab it also appears as two sibling rows — `left-pad  (1.0.0 → 1.1.0)` and `left-pad  (details)` — and a user will reasonably ask whether that's one change or two.

The parallel classification is right for JSON, but for human-facing *totals* I'd either (a) count distinct occurrences (transition-only rows add to the total; version+transition rows count once), or (b) fold the transition evidence into the existing "changed" row's detail pane when the same occurrence has both, keeping standalone rows only for transition-only changes. Option (b) is the better UX: one package, one row, all its deltas together.

internal/mcp/compact_diff.go
```diff
@@ -17,33 +21,53 @@ type SecurityDelta struct {
...
-	Informational []CompactFinding      `json:"informational,omitempty"`
-	Diagnostics   []Diagnostic          `json:"diagnostics,omitempty"`
-	Truncation    *TruncationInfo       `json:"truncation,omitempty"`
-	Hint          string                `json:"hint,omitempty"`
+	SchemaVersion string                        `json:"schema_version"`
+	Command       string                        `json:"command"`
+	Comparison    output.DiffComparison         `json:"comparison"`
+	Summary       CompactDiffSummary            `json:"summary"`
+	SecurityDelta SecurityDelta                 `json:"security_delta"`
+	Transitions   []CompactDependencyTransition `json:"dependency_transitions,omitempty"`
```
bomly-guy: Naming consistency for agent consumers: the CLI JSON calls this list `transitions`, the MCP response calls it `dependency_transitions`, and the summary counter is `packages_transitioned`. An agent moving between `bomly diff --format json` and the `bomly_diff` tool has to learn three names for one concept. Suggest aligning on `transitions` here (the `Compact` prefix and the response context already scope it), and `packages_with_detail_changes`—or at minimum the same root word—for the counter. The `diffHint` sentence would then also read more naturally: it currently says "Dependency transitions describe…" which matches none of the human-facing labels.

internal/cli/render/diff_markdown.go
```diff
@@ -56,7 +56,13 @@ func diffOverviewMarkdown(payload output.DiffResponse) []string {
 		[][]string{{
 			status,
 			fmt.Sprintf("+%d / ~%d / -%d", payload.Summary.AddedManifestCount, payload.Summary.ChangedManifestCount, payload.Summary.RemovedManifestCount),
-			fmt.Sprintf("+%d / ~%d / -%d", payload.Summary.AddedPackageCount, payload.Summary.ChangedPackageCount, payload.Summary.RemovedPackageCount),
+			fmt.Sprintf(
+				"+%d / ~%d version / ~%d details / -%d",
```
bomly-guy: Wording nit: `+2 / ~3 version / ~1 details / -0` is dense for a first-time reader of a PR comment (Guard's main surface). Since this cell has room, consider spelling it out: `+2 added / 3 version / 1 details / -0 removed` — or keep the compact form but make the column header carry the legend (`Packages (+add / ~ver / ~details / -rem)`). Guard PR comments are read by people who never ran the CLI, so the markdown surface should be the *most* self-explanatory, not the tersest.

bomly-guy (commented):
## Review: correctness, architecture, and — mainly — whether "transitioned dependencies" is a clear, useful concept

**Verdict:** the signal is genuinely valuable and the architecture is right (canonical classifier in the SDK, output layers only project, schemas stay additive, targeted tests pass). But the PR ships with one likely CI breaker, one real-world noise-flood scenario, and — on the concept itself — the naming is fragmented across surfaces and the `registry_eligibility` field is usually redundant with `source`, so one conceptual change gets reported as two or three. Details are in the inline comments; summary below.

### Is the concept clear and useful?

**Useful — arguably more useful than the PR presents it.** direct↔transitive changes affect remediation strategy and policy; registry→git source changes are a provenance/supply-chain red flag; losing registry-matching eligibility means *"Bomly can no longer check this package against vulnerability databases"* — a security-coverage regression users absolutely want surfaced in a PR diff. Previously these were invisible or misread as add/remove pairs.

**But three clarity problems keep it from landing:**

1. **Fragmented vocabulary.** The same thing is called `transitions` (JSON), `dependency_transitions` / `packages_transitioned` (MCP), "Dependency detail changes" (text/markdown), "Details changed" (TUI counts), a `TRANSITIONED` badge (TUI), and "metadata transitions" (docs/SDK). Pick **one** human-facing term and use it everywhere. Recommendation: keep **"Detail changes"** as the short human label (paired with "Version changed"), keep `transitions` as the stable JSON noun, and kill "transitioned" as a user-visible word — on its own it tells a user nothing (transitioned to *what?*).
2. **The eligibility field states mechanism, not consequence — and is usually redundant.** `RegistryMatchEligible()` is derived almost entirely from `Source`, so the common case (registry → git) reports *three* changed fields for one conceptual change — this PR's own test asserts exactly that. A user reads "source: registry → git; registry matching: eligible → not eligible" as the same fact twice, the second in internal jargon. Suggested fix: in human-readable surfaces, only show the eligibility line when it changed *independently* of source, and reword it to the consequence — **"vulnerability check coverage: covered → no longer covered"**. Keep the raw structured fields in JSON/MCP as-is.
3. **Flat severity framing.** All transitions render identically (neutral cyan), but they aren't equally important: `direct → transitive` is informational, while `registry → git` or losing enrichment coverage deserves warn-level emphasis and is a natural future input to `--audit` policy. Right now the most security-relevant new signal in the product is styled as the least interesting row on the screen.

### Blocking before merge

- **Smoke goldens will drift** (`transitioned_package_count` has no `omitempty`; six diff goldens contain the summary block and are not regenerated here — see inline comment). The feature checklist also calls for a smoke case that meaningfully exercises the feature; no smoke case currently produces a transition.
- **Cross-detector diffs flood the report** (SBOM detector never sets `Source`; lockfile detectors do — see inline comment on `CompareDependencyMetadata`).

### Fast-follow (non-blocking)

- Eligibility de-duplication + consequence-oriented wording (inline comments).
- Double-count presentation in TUI aggregates (inline comment).
- Relationship-inference perf hoist for legacy graphs (inline comment).
- `DependencyMetadataTransition` naming collision with `Dependency.Metadata` (inline comment).

Verified locally at the PR head: `go test ./sdk/... ./internal/output/... ./internal/mcp/... ./internal/tui/... ./internal/cli/render/...` — all green.

test/smoke/smoke_test.go
```diff
@@ -384,6 +386,11 @@ func TestDiff(t *testing.T) {
 			name: "diff-sbom",
 			args: []string{"diff", "--sbom", "--base", sbomFixture("go.spdx.json"), "--head", sbomFixture("js.spdx.json"), "--format", "json"},
 		},
+		{
+			name:                "diff-sbom-detail-change",
```
bomly-guy, line 390: **Non-blocking nit — CI slice ownership is implicit.** Both workflow slice matrices match this subtest only via the unanchored pattern `TestDiff/diff-sbom` (`.github/workflows/smoke.yml:97`, `.github/workflows/update-smoke-goldens.yml:121`), which happens to also match `diff-sbom-detail-change`. That works today — exactly one slice runs it and produces its golden — but it's fragile in exactly the way CLAUDE.md warns about: if someone later "fixes" the pattern to `diff-sbom$` for exactness, this golden silently drops out of regeneration with no gap detection. Suggest making ownership explicit in both workflows:

```
TestDiff/(diff-sbom$|diff-sbom-detail-change$)
```

Fine to do here or as a trivial follow-up.

internal/cli/render/diff.go
```diff
@@ -65,15 +65,81 @@ func dependencyTextSections(results output.DiffDependencyResults) []string {
 			lines = append(lines, Wrap(line, Yellow))
 		}
 	}
+	appendTransitions := func() {
+		if len(results.Transitions) == 0 {
+			return
+		}
+		lines = append(lines, Style(fmt.Sprintf("Detail changes (%d)", len(results.Transitions)), Bold))
```
bomly-guy, line 72: **Non-blocking, deferred fast-follow — severity framing.** All detail changes still render with equal weight, but they aren't equally important: `relationship: direct → transitive` is informational, while `source: registry → git` is a provenance/supply-chain red flag and a coverage loss ("vulnerability checks: covered → not covered") is a security regression. Two future steps worth tracking as an issue so this doesn't get lost:

1. Warn-level styling (yellow/⚠) for source changes *to* `git`/`url` and for coverage losses, across text, markdown, and TUI.
2. Feed detail transitions into `--audit` as policy input — e.g. a policy that fails when a dependency's source leaves `registry`. That would turn this from a report row into an enforceable control, which is where the real value of this signal lands.

bomly-guy (commented):
## Re-review of b34b4cc — all previous review points resolved

The follow-up commit addresses every point from the earlier review, including both blockers. Verified hands-on at this head: `go build ./...` passes, all targeted test packages pass (`sdk`, `internal/output`, `internal/mcp`, `internal/tui`, `internal/cli/render`, `internal/cli`), the new `TestDiff/diff-sbom-detail-change` smoke subtest passes end-to-end locally with the `smoke` tag, and `make generate` leaves the tree clean. **I'd consider this mergeable.**

Point-by-point:

- **Smoke goldens (was blocking)** — ✅ all six diff goldens regenerated, plus a new `diff-sbom-detail-change` smoke case built on SBOM fixtures (the one testdata type the repo allows). Good test design: it asserts the JSON contract (`transitioned_package_count`, `results.dependencies.transitions`) *and* re-runs in text format to assert the human wording `relationship: direct → transitive`.
- **Cross-detector noise flood (was blocking)** — ✅ fixed as suggested and then some: relationship changes are only reported when *neither* side is `unknown`, source changes only when *both* sides are non-empty, and eligibility only when the source evidence is comparable (equal, or both present). SBOM-vs-lockfile diffs go quiet by default, while a genuine first-party flip with matching sources still reports.
- **Eligibility redundancy** — ✅ human surfaces (text and TUI) skip the eligibility line whenever `source` also changed, and the wording is now "vulnerability checks: covered → not covered". JSON/MCP keep the raw structured fields. This was the biggest UX concern and it's resolved cleanly.
- **Naming unification** — ✅ `DependencyDetailTransition` in the SDK (the `Metadata` collision is gone), "Detail changes" consistently across text, markdown, and TUI, badge changed from `TRANSITIONED` on neutral to `DETAILS` on yellow (also fixing the color collision with `unchanged`), MCP renamed to `transitions` + `packages_with_detail_changes`, `diffHint` updated to match. Renaming the MCP field pre-merge is free since `dependency_transitions` never shipped.
- **Double counting** — ✅ and the better option was taken: a package with both a version and a detail change is now **one** row ("Version and dependency detail changes") with the transition evidence folded into its detail pane, and all aggregate counts (`PackageDeltas`, ecosystem histogram, top-changed manifests) count distinct occurrences via before/after pair keys. Standalone rows remain only for detail-only changes.
- **Perf** — ✅ relationship inference is now a single precomputed per-graph map instead of per-node root walks, consolidated with `RelationshipForPath` into a shared `relationshipForDepth` helper — nicer than what was asked for.
- **Markdown legend** — ✅ the cryptic `+2 / ~3 version / ~1 details / -0` is now spelled out (`2 added / 3 version changed / 1 detail changes / 0 removed`), with pluralization handled.

Two tiny non-blocking leftovers, both fine to ship as-is — see the inline comments:

1. The CI slice pattern `TestDiff/diff-sbom` is unanchored and only *happens* to cover the new subtest (inline comment on `smoke_test.go`).
2. The severity-framing suggestion (warn-level emphasis for `registry → git`, transitions as `--audit` policy input) remains an explicitly deferred fast-follow (inline comment on `render/diff.go`).

PR #348, opened 2026-07-28
Review and audit risky dependency detail changes

bomly-guy:
## Summary

- mark dependency source moves to Git or URLs for review in text, Markdown, and the interactive diff view
- explain that registry-based vulnerability checks may no longer cover the dependency
- pass canonical head-side detail transitions to the existing package auditor
- make source changes fail with the shared `--fail-on source-change` policy
- keep CLI schema `1.0`, MCP schema `mcp/1`, and plugin protocol `v1`

## Design

This keeps one canonical transition model. Presentation uses the shared SDK classifier, while the existing package auditor owns source-change policy:

- source changes to Git or an arbitrary URL warn by default
- `--fail-on source-change` makes either source change fail
- `--warn-only` can still downgrade the failure
- vulnerability severity, reachability, and exploitability constraints keep their existing AND behavior; `source-change` is an independent package-finding gate
- using `source-change` alone leaves the normal vulnerability inventory unchanged

The earlier inferred coverage-loss finding has been removed. Registry eligibility describes whether a dependency can be sent to registry matchers; it does not prove that matching was attempted or failed. Source-change warnings still explain the possible coverage effect without creating a second finding.

Relationship-only changes, registry-eligibility-only changes, gains, and missing evidence remain informational. Only Git and URL moves are review signals; project, workspace, and file moves remain informational. No new auditor or pipeline stage is added. Source-change checks currently rely on evidence from npm, pnpm, Yarn, and Bun detectors; broader detector source classification is deferred.

The safe `--audit requires --enrich` default remains unchanged. A possible explicit existing-evidence-only audit mode is tracked separately in #351.

## Public fixture

The npm smoke case compares two immutable annotated tags in `bomly-dev/example-javascript-npm`. Both contain `minimist@1.2.8`; the head changes only its source from the npm registry to a pinned Git commit. The repository default branch is unchanged, the same-named mutable fixture branch has been removed, and the smoke case disables all external matchers.

## Commits

1. `Highlight risky dependency detail changes`
2. `Audit risky dependency detail changes`
3. `Address dependency detail review feedback`
4. `Complete dependency detail review coverage`
5. `Default source-change enforcement to any`
6. `Simplify dependency source change policy`
7. `Preserve vulnerability findings with source policy`

## Validation

- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `go test -tags smoke ./test/smoke -run 'TestDependencyDetailRiskPolicy$' -count=1 -v`


internal/auditors/vulnerability/auditor.go
```diff
@@ -100,6 +102,91 @@ func (a Auditor) Audit(_ context.Context, req sdk.AuditRequest) (sdk.AuditResult
...
+func transitionHasReviewReason(transition sdk.DependencyDetailTransition, wanted sdk.DependencyDetailReviewReason) bool {
+	for _, reason := range transition.ReviewReasons() {
+		if reason == wanted {
+			return true
+		}
+	}
+	return false
+}
+
+func coverageLossMatchesConstraints(constraints []sdk.FailOnConstraint) bool {
```
bomly-guy: **This is the main change I'd like before merge** (detail in the review summary).

Deriving coverage-loss enforcement from the severity ladder means the rule can only be enabled by lowering the vulnerability threshold to `medium` — and it is silently unavailable to anyone running `--fail-on high`, which is the most common CI setting. It also can't be combined with `reachable`/`exploitable`, because the AND-walk here returns `false` the moment a non-severity constraint appears; `--fail-on medium --fail-on exploitable` quietly downgrades coverage loss back to a warning with no way to notice.

Please give it its own token instead: `--fail-on coverage-loss` (a new `CoverageConstraint` kind in `sdk/policy.go`). Then this function matches on the token, `--fail-on coverage-loss --fail-on high` composes the way a user would expect, and the rule stops being an accident of `SeverityRank(warning) == SeverityRank(medium)`.

I'd document the token as `coverage-loss` rather than `loss` — `loss` doesn't say loss *of what*, and the existing token vocabulary is self-describing. A hidden `loss` alias is fine if you want the shorthand.

internal/auditors/vulnerability/auditor.go
```diff
@@ -100,6 +102,91 @@ func (a Auditor) Audit(_ context.Context, req sdk.AuditRequest) (sdk.AuditResult
...
+		}
+		status := sdk.FindingPolicyStatusWarn
+		if coverageLossMatchesConstraints(constraints) {
+			status = sdk.FindingPolicyStatusFail
+		}
+		grouped[identity] = &sdk.Finding{
+			ID:             fmt.Sprintf("%s:%s:%s", auditorName, ruleID, identity),
+			Kind:           sdk.FindingKindPackage,
+			Title:          "Vulnerability checks no longer cover dependency",
+			Severity:       sdk.SeverityWarning,
```
bomly-guy: Severity is pinned to `SeverityWarning` even when `PolicyStatus` resolves to `fail` on line 125-127. The package auditor's new source-change rule does the opposite — `packageFindingSeverity(status)` promotes a failing finding to `SeverityError`.

So two rules added in the same PR disagree on whether a failing finding is `warning` or `error`, and that difference is user-visible: SARIF `security-severity` and the GitHub code-scanning annotation level are derived from `Severity`, so a coverage-loss finding that fails the build still renders as a warning annotation in Guard.

Pick one. Given `--fail-on` reads severity today, hardcoding `warning` here is also what forces the `SeverityRank(warning) == medium` coupling in the comment above; a dedicated `coverage-loss` token would let you use `packageFindingSeverity(status)` here without any circularity.

docs/auditors/package.md
```diff
@@ -2,36 +2,39 @@
...
 | Needs enrichment data | No |
-| Policy flags | `--deny-package`, `--deny-group`, `--protected-package`, `--typosquat-threshold`, `--typosquat-mode` |
-| Finding reasons | denied package, denied group, typosquat of protected package |
+| Policy flags | `--deny-package`, `--deny-group`, `--protected-package`, `--typosquat-threshold`, `--typosquat-mode`, `--deny-dependency-source-change` |
+| Finding reasons | denied package, denied group, typosquat of protected package, source changed to Git or URL |
 
 ## User notes
 
-Name-based, so it needs no enrichment and runs fully offline. Declare the packages you trust with `--protected-package`; Bomly flags lookalikes within `--typosquat-threshold`. `--typosquat-mode` sets finding policy status, while `--fail-on` controls whether a matching finding changes the exit code.
+Package identity checks need no enrichment. In an enriched diff audit, the auditor also warns when a known dependency source changes to Git or a URL. Use `--deny-dependency-source-change` when that change should fail.
```
bomly-guy: Worth stating the ecosystem scope explicitly here and in `docs/auditors/vulnerability.md`.

`Dependency.Source` is only populated by the Node detectors today — `internal/detectors/node/{npm,pnpm,yarn,bun}` via `DependencySourceFromSpecifier`. Every other detector leaves it empty, and `compareDependencyDetails` requires `before.Source != "" && after.Source != ""` before it records either `DependencyDetailSource` or `DependencyDetailRegistryEligibility`. So on a Go, Python, Cargo, Composer, Maven, or Gradle repo, `--deny-dependency-source-change git` is accepted, validated, logged, and can never fire.

That's a silent no-op on the majority of supported ecosystems — exactly the shape a security policy flag should not have. These are all ecosystems where the risky move exists in practice (`replace` directives and VCS pseudo-versions in Go, `git+https://` requirements in Python, `git = ` in Cargo, `vcs` repositories in Composer). Two things:

1. Add a **Limitations** line naming npm/pnpm/yarn/bun as the current coverage.
2. Consider a follow-up issue for populating `Source` in the other detectors, so the flag's blast radius grows to match its promise.

test/smoke/audit_test.go
```diff
@@ -263,3 +267,93 @@ func TestAuditDiffAndExplain(t *testing.T) {
...
 	}
 }
+
+func TestDependencyDetailRiskPolicy(t *testing.T) {
+	requireTool(t, "npm")
+	common := []string{
+		"diff",
+		"--url", "https://github.com/bomly-dev/example-javascript-npm",
+		"--base", "assurance/dependency-source-registry-v1",
+		"--head", "assurance/dependency-source-git-v1",
```
bomly-guy, line 277: The PR body says both sides are immutable refs, but `assurance/dependency-source-git-v1` exists in `bomly-dev/example-javascript-npm` as **both** a tag and a branch:

```
455a3ec  refs/heads/assurance/dependency-source-git-v1
96ffda2  refs/tags/assurance/dependency-source-git-v1
455a3ec  refs/tags/assurance/dependency-source-git-v1^{}
```

They resolve to the same commit right now, but the ref name is ambiguous and the branch is mutable, so a stray push to that branch silently changes what the smoke case tests. (`assurance/dependency-source-registry-v1` is tag-only — correct.)

Easiest fix: delete `refs/heads/assurance/dependency-source-git-v1` in the fixture repo so only the tag remains. CLAUDE.md's smoke rule ("pinned to a specific tag or commit SHA") is satisfied either way, but not while a same-named branch shadows the tag.

internal/cli/opts/flag_options.go
```diff
@@ -60,6 +63,15 @@ func BindCommandFlagGroups(cmd *cobra.Command, cfg *config.Resolved, groups ...F
 	return nil
 }
 
+func bindDiffPolicyFlags(flags *pflag.FlagSet, cfg *config.Resolved) {
+	flags.StringArrayVar(
+		&cfg.DenyDependencySourceChanges,
+		"deny-dependency-source-change",
```
bomly-guy: The flag is diff-only (`FlagGroupDiffPolicy` is bound to `diffCmd` alone), but the config surface is global: `policy.deny_dependency_source_changes` in a YAML file is read on every command, wired through `RegistryConfigsFromResolved` into the package auditor, and then does nothing on `scan` because scan never populates `DependencyDetailChanges`.

So the natural workflow — put the policy in `.bomly.yaml`, run `bomly scan --enrich --audit` in CI — passes validation and produces zero findings, with nothing telling the user the key doesn't apply. Either warn when the key is set on a non-diff command, or say "diff-only" in the `doc:` tag on `Resolved.DenyDependencySourceChanges` so it lands in the generated `docs/CONFIG_REFERENCE.md` row (the options table in `docs/auditors/package.md` says it, but the config reference doesn't).

Separately: gating this on `--audit`, which itself requires `--enrich`, means a purely local structural check (base lockfile said `registry`, head lockfile says `git` — no network data involved) can't run without enrichment. Not blocking, but it makes the cheapest and arguably highest-signal check in this PR the one with the most expensive precondition.

internal/output/view.go
```diff
@@ -1124,6 +1124,27 @@ func diffDependencyTransitionsFromDiff(transitions []sdk.DependencyDetailTransit
 	return out
 }
 
+// DependencyDetailReviewReasons classifies a projected dependency detail
+// transition without adding a derived field to the JSON contract.
+func DependencyDetailReviewReasons(transition DiffDependencyTransition) []sdk.DependencyDetailReviewReason {
```
bomly-guy, line 1129: This reconstructs a synthetic `sdk.DependencyDetailTransition` from the projected view state so presentation and audit share one classifier — good instinct, and better than duplicating the rules. But the projection is lossy by construction: it only carries `Source`, `ChangedFields`, relationship, and eligibility. If `ReviewReasons()` ever grows a reason that reads anything else on `Before`/`After` (name, version, PURL, first-party flag), presentation silently stops flagging it while the auditor keeps failing the build — a divergence with no compile error and no test failure.

A parity test would close this cheaply: build one real transition, run it through `sdk.ReviewReasons()` and through `diffDependencyTransitionsFromDiff` → `output.DependencyDetailReviewReasons`, assert the two slices are equal. Then any future field dependency breaks loudly here instead of quietly in a diff report.

bomly-guy (commented):
Overall this is a clean, well-scoped change: one canonical classifier in `sdk`, presentation reads it, audit policy stays in the existing auditors, no new pipeline stage. Tests are thorough (the policy matrix and the clone-per-auditor test are good). I ran `go test ./sdk/... ./internal/auditors/... ./internal/config/... ./internal/engine/... ./internal/cli/render/... ./internal/tui/...` on the branch — all green.

Main request, plus a handful of smaller notes inline.

### Coverage loss should be its own `--fail-on` rule, not a severity side effect

Right now `vulnerability-coverage-loss` is emitted at `warning` severity, and `SeverityWarning` ranks equal to `medium`, so `--fail-on medium` (or `low`/`any`) turns it into a build failure and `--fail-on high`/`critical` does not. That couples two unrelated decisions:

- A team that only wants to fail on high/critical CVEs *cannot* enforce coverage loss at all.
- A team that wants to enforce coverage loss has to drop its vulnerability threshold to `medium`, which pulls in every medium advisory as a side effect.

Coverage loss is not an advisory severity — it is a supply-chain integrity signal. It should get its own token alongside `reachable` and `exploitable`:

```bash
bomly diff --base main --head HEAD --enrich --audit --fail-on coverage-loss
```

I'd go with the explicit `coverage-loss` over a short `loss` — `--fail-on loss` doesn't say loss *of what*, and the token list is already read as a vocabulary (`any|low|medium|high|critical`, `reachable`, `exploitable`), so a self-describing token fits better and stays greppable in CI configs. If brevity matters we can accept `loss` as a hidden alias, but the documented spelling should be `coverage-loss`.

That means a new `CoverageConstraint` kind in `sdk/policy.go`, and `coverageLossMatchesConstraints` matching on that token instead of walking severity ranks. Note this also cleanly fixes the AND semantics: today `--fail-on medium --fail-on exploitable` silently downgrades coverage loss back to a warning because an integrity signal can never satisfy `exploitable`. With a dedicated token, `--fail-on coverage-loss --fail-on high` reads exactly as intended — fail on high advisories *and* on coverage loss.

Same reasoning arguably applies to the two source-change rules, which already have their own flag (`--deny-dependency-source-change`) rather than riding severity — so this change would make the three new rules consistent with each other.

sdk/policy.go
```diff
@@ -136,33 +152,45 @@ func SeverityMeets(candidate SeverityLevel, threshold string) bool {
...
+			// Source changes are evaluated by the package auditor against
+			// dependency detail transitions, not vulnerabilities.
 		default:
 			// Unknown kinds are treated as no-op rather than as
 			// rejection so future constraint kinds can be added without
 			// breaking older auditor behavior.
 		}
 	}
-	return true
+	return evaluated
```
bomly-guy: **Blocking** — see the review summary for the full write-up and the empirical run.

Switching the tail from `return true` to `return evaluated` means a constraint list with no vulnerability-kind entry matches nothing. Because the vulnerability auditor gates finding *creation* on this predicate (`internal/auditors/vulnerability/auditor.go:76`), `--fail-on source-change` doesn't add a gate — it deletes every advisory from the report and exits 0.

The cleanest fix is to keep this returning `true` when no vulnerability constraint was supplied. "The user named zero vulnerability constraints" and "the user named vulnerability constraints that this advisory failed" are different states, and only the second should suppress a finding. Something like:

```go
for _, c := range constraints {
    switch c.Kind {
    // ... unchanged, minus the `evaluated` bookkeeping
    }
}
return true
```

leaves `--fail-on high` filtering exactly as before, and lets `--fail-on source-change` mean "advisories as usual, plus fail on source changes" — which is how the flag reads and how the docs describe combining it.

sdk/policy.go
```diff
@@ -136,33 +152,45 @@ func SeverityMeets(candidate SeverityLevel, threshold string) bool {
...
 		case ExploitabilityConstraint:
+			evaluated = true
 			if !v.IsExploitable() {
 				return false
 			}
+		case SourceChangeConstraint:
+			// Source changes are evaluated by the package auditor against
+			// dependency detail transitions, not vulnerabilities.
 		default:
 			// Unknown kinds are treated as no-op rather than as
```
bomly-guy, line 186: This comment now says the opposite of what the code does. The stated contract is that unknown kinds are a no-op "so future constraint kinds can be added without breaking older auditor behavior" — but an unknown kind no longer sets `evaluated`, so a list containing only a future/plugin constraint kind rejects every vulnerability rather than passing them through.

That's precisely the forward-compat break the comment promises won't happen, and it's silent: an older binary reading a newer config drops all advisories instead of ignoring the token it doesn't understand. Fixing the `return evaluated` issue above resolves this too; otherwise the comment needs to be rewritten to describe the new (much sharper) behavior.

docs/AUDITORS.md
```diff
@@ -67,8 +74,12 @@ The `any` token matches every severity, including `unknown`.
 | `critical` | findings with severity = critical |
 | `reachable` | findings where reachability status is `reachable` (experimental — see [REACHABILITY.md](REACHABILITY.md)) |
 | `exploitable` | vulnerability findings marked as known exploited by enrichment data |
+| `source-change` | diffs where a dependency changes from a known source to Git or an arbitrary URL |
 
-Repeat the flag to AND constraints together:
+Repeat vulnerability constraints to AND them together. `source-change`
+is an independent package gate, so combining it with vulnerability constraints
```
bomly-guy, line 80: "combining it with vulnerability constraints fails when either the source gate or the complete vulnerability constraint set matches" is accurate for the *combined* case, but the doc never covers `source-change` on its own — which is the spelling a user is most likely to reach for and the one that currently zeroes out advisory reporting.

If the semantics change per the blocking comment, this paragraph is fine as-is. If they don't, this table and the token row on line 77 need to say outright that `source-change` alone suppresses vulnerability findings and must be paired with a severity token (`--fail-on any --fail-on source-change`) to keep them.

Also worth a line here or in `docs/auditors/package.md`: `source-change` is accepted on `bomly scan`, where it can never match (scan produces no detail transitions). Right now it's silently inert there.

docs/auditors/package.md
```diff
@@ -2,36 +2,40 @@
...
 | Needs enrichment data | No |
-| Policy flags | `--deny-package`, `--deny-group`, `--protected-package`, `--typosquat-threshold`, `--typosquat-mode` |
-| Finding reasons | denied package, denied group, typosquat of protected package |
+| Policy flags | `--deny-package`, `--deny-group`, `--protected-package`, `--typosquat-threshold`, `--typosquat-mode`, `--fail-on source-change` |
+| Finding reasons | denied package, denied group, typosquat of protected package, source changed to Git or URL |
 
 ## User notes
 
-Name-based, so it needs no enrichment and runs fully offline. Declare the packages you trust with `--protected-package`; Bomly flags lookalikes within `--typosquat-threshold`. `--typosquat-mode` sets finding policy status, while `--fail-on` controls whether a matching finding changes the exit code.
+Package identity checks need no enrichment. In a diff audit, the auditor also warns when a known dependency source changes to Git or a URL. Use `--fail-on source-change` when that change should fail.
```
bomly-guy, line 18: Not blocking — a question about the scope reduction, since dropping the coverage-loss finding changed what the PR delivers relative to its title.

The reasoning in the description is sound: `RegistryMatchEligible()` says a dependency *can* be sent to registry matchers, not that matching was attempted or failed, so a `vulnerability-coverage-loss` finding was inferring an outcome from a precondition. Folding the caveat into the source-change finding title and the rendered description is more honest, and #351 is the right home for an evidence-based version.

The residual gap: `ReviewReasons()` now fires only for `git` and `url`, but `RegistryMatchEligible()` also returns false for `project`, `workspace`, and `file`. A `registry → file:` transition is a real eligibility loss that now gets no review marker, no finding, and no prose — it renders as `Info` alongside a relationship change. Most of those are benign monorepo refactors, which is presumably why they're excluded, but `file:` pointing at a vendored tarball is the same scanner-blinding shape as `git:` and is arguably *less* auditable (no upstream to compare against).

Worth either extending the reasons to cover it, or stating explicitly in Limitations that eligibility loss is only surfaced for Git and URL sources — so a reader doesn't infer that an unmarked detail change means coverage was retained.

bomly-guy (commented):
Re-reviewed at `fa19e00`. Everything from the last round is addressed, and the pivot to `--fail-on source-change` is a better design than the dedicated `--deny-dependency-source-change` flag it replaced — one policy vocabulary, no diff-only flag with a global YAML key, and severity now derives from `packageFindingSeverity(status)` consistently. The fixture branch is gone (`git ls-remote` now shows tags only), the detector scope is documented as a limitation, and the projection-parity test is in place. `make test` is green on the branch.

**One blocking issue** — introduced by the `MatchesConstraints` rewrite, not present in the previous revision. Please don't merge as-is.

### `--fail-on source-change` silently disables all vulnerability reporting

`MatchesConstraints` now returns `evaluated` instead of `true`, so a constraint list containing *only* non-vulnerability kinds matches no vulnerability. The vulnerability auditor uses that same predicate to decide whether to **create** a finding at all (`auditor.go:76` — `!MatchesConstraints(...)` → `continue`), not merely whether it fails. So `source-change` used alone doesn't add a gate; it removes every advisory from the report.

I confirmed this against the branch with a throwaway test — one package, one critical CVE, four policies:

```
--fail-on (none)                => 1 critical vulnerability findings
--fail-on high                  => 1 critical vulnerability findings
--fail-on source-change         => 0 critical vulnerability findings   <--
--fail-on high + source-change  => 1 critical vulnerability findings
```

Nothing rejects or warns about this. `config.Validate` doesn't constrain `source-change`, it's a global flag rather than diff-scoped, and `render.Scan` gives no hint. The failure mode:

```bash
# reads as "also fail when a dependency's source changes"
bomly diff --base main --head HEAD --enrich --audit --fail-on source-change
```

A reviewer expects source-change enforcement *plus* the usual advisory reporting. What they get is source-change enforcement and a diff that reports zero introduced vulnerabilities — green build, critical CVE shipped. It's worse on `scan`, where `--fail-on source-change` can never fire at all (no transitions) yet still suppresses everything: a pure `--audit` downgrade that looks like a policy tightening. This is easy to reach by copying a policy line between a diff job and a scan job, and the more security-conscious the user, the more likely they are to add the token.

The asymmetry is the root cause: severity/reachability/exploitability are *filters that narrow* a finding set, `source-change` is a *gate that adds* one, and they now share a flag where repeating means AND for one group and OR across groups. Options, roughly in order of preference:

1. Leave `MatchesConstraints` returning `true` for a list with no vulnerability constraints (i.e. treat "no vulnerability constraint specified" as the existing unfiltered default) and let the package auditor's independent gate do its own thing. `--fail-on source-change` then means "report advisories as usual, and additionally fail on source changes" — which is what the flag reads like.
2. Keep the current semantics but reject `source-change` as the sole `--fail-on` value in `config.Validate`, with a message pointing at option 1's spelling (`--fail-on any --fail-on source-change`).
3. Give source changes their own flag again, outside `--fail-on`.

Whichever you pick, please add a test that pins advisory visibility under `--fail-on source-change` — the current suite covers the constraint parsing and the package gate, but nothing asserts what happens to vulnerability findings when the token appears alone.

PR #352, opened 2026-07-29
Expand dependency source classification

bomly-guy:
## Summary

- classify dependency origins from explicit Cargo, Bundler, pub, SwiftPM, and Python detector evidence
- preserve available Git revisions while leaving missing or unrecognized evidence unknown
- reuse the existing diff review and package-audit behavior without changing schemas, policy flags, or plugin protocols
- document the exact supported detector scope and the formats that cannot yet prove origin

## Why

Source-change review only works when both sides of a diff contain reliable occurrence-level source evidence. Several lockfiles and tool outputs already provide that evidence, but their detectors were not mapping it to the shared dependency source model.

The classification stays detector-owned. Central diff and audit code consumes the shared value and does not guess from package names, ecosystems, PURLs, or repository metadata.

## Matching behavior

Source classification also controls which occurrences are sent to external package matchers. Registry releases remain eligible. Git, file, and arbitrary URL occurrences are normally excluded because they may not represent the published registry release.

SwiftPM remote source-control packages are the deliberate exception. Their repository URL is the canonical SwiftURL package identity used by vulnerability sources, so they remain eligible for vulnerability matching while retaining truthful `git` source evidence. Local Swift packages remain ineligible.

## User impact

`bomly diff` can now review registry-to-Git or registry-to-URL changes for Cargo, Bundler, pub, SwiftPM, pip, Pipenv, Poetry, and uv projects in addition to the JavaScript package managers already covered. Unsupported or malformed source evidence remains silent rather than producing a false warning.

## Validation

- `make generate`
- `make fmt-check`
- `make lint`
- `make test`
- `make build`
- `git diff --check`


bomly-guy (commented):
## Review

**Overview.** Extends detector-owned `sdk.Dependency.Source` classification from the JS package managers to Cargo, Bundler, pub, SwiftPM, and the four Python paths (pip inspect, Pipenv, Poetry, uv), preserving Git revisions in `source_revision` metadata where the format proves them. Central diff/audit consumers are untouched, and I verified the claims in the PR body hold on the consumer side: `compareDependencyDetails` requires both sides non-empty before a source change fires, and an empty source stays registry-eligible. The evidence-only stance (unknown stays silent) is consistently applied and well tested at the classifier level. Docs are updated in both `docs/` and `internal/support/prose/` mirrors, plus the architecture decision log — nice.

**However, the PR body understates the blast radius.** Classification is not diff-only: `RegistryMatchEligible()` returns false for `git`/`file`/`url` sources, and `engine.registryMatchRequest` uses it to build the matcher-facing graph. So every package this PR classifies as Git/File/URL is silently removed from OSV/Grype/deps.dev/scorecard enrichment. For Cargo git pins, Bundler `GIT` gems, and Python direct-URL installs that follows the model established for npm in #347/#348 and is arguably correct — but it deserves a callout in the PR body and user docs. For **SwiftPM it is a regression that effectively disables vulnerability enrichment for the whole ecosystem** (inline comment on `swiftpm/detector.go`).

### Findings

1. **[blocking] SwiftPM enrichment regression** — `remoteSourceControl` (and legacy `repositoryURL`) → `DependencySourceGit` makes essentially every SwiftPM dependency ineligible for registry matching. See inline comment.
2. **[medium] Cargo workspace-from-lock path classifies nothing** — `depGraphFromLockWorkspace` in `internal/detectors/cargo/workspace.go` builds nodes with a local `nodeFor` closure that never sets `Source`, even though `parseCargoLockPackages` now extracts it. See inline comment on `cargo/detector.go`.
3. **[question] uv revision keys may not exist in real lockfiles** — see inline comment on `uvlock.go`.
4. **[minor] style nits** — inline comments on `cargo/detector.go` and `ruby/detector.go`; also `firstNonEmptyPython` in `poetrylock.go` doesn't need the `Python` suffix (there is no `firstNonEmpty` in package `python` to collide with — swiftpm's is a different package).

### Test coverage

Classifier-level table tests are thorough (good negative cases: unknown Cargo scheme, pub `sdk`, unknown SwiftPM pin kind, Bundler `PLUGIN` section). Gaps worth closing:

- No test covers the enrichment interaction — e.g. an assertion that a SwiftPM remote package remains (or intentionally stops being) registry-match-eligible. Given finding 1, whichever way it resolves should be pinned by a test.
- Lockfile integration tests only assert the registry happy path (`requirePySource(..., requests, registry)`); a git-pinned entry in one Python fixture would exercise the interesting branch end-to-end.
- No smoke case exercises a registry→git transition for the newly covered ecosystems (`diff-npm` is the only detail-change diff smoke). Fine as a follow-up.

### Goldens / output

I checked the output projections: scan JSON serializes neither `source` nor `resolved_url` on packages/dependencies, and node-level metadata doesn't reach the document either, so existing scan smoke goldens should not drift from this change (Bundler's new `ResolvedURL = rubygems remote` and uv's registry URL stay internal). Diff transitions only appear on actual source changes. The scorecard matcher's repo resolution now also sees Bundler/uv URLs — harmless for rubygems/pypi (only GitHub URLs are extracted), and a small win for git-pinned gems.

internal/detectors/swiftpm/detector.go
```diff
@@ -276,6 +279,39 @@ func packageNode(pkg swiftPackage) *sdk.Dependency {
 	return node
 }
 
+func swiftDependencySource(kind, location string) sdk.DependencySource {
+	switch strings.ToLower(strings.TrimSpace(kind)) {
+	case "registry":
+		return sdk.DependencySourceRegistry
+	case "remotesourcecontrol":
```
bomly-guy, line 286: **[blocking] This disables vulnerability enrichment for essentially all SwiftPM projects.**

`RegistryMatchEligible()` returns false for `DependencySourceGit`, and `engine.registryMatchRequest` filters ineligible nodes out of the graph handed to matchers. Swift package adoption of the registry is still rare — in practice every SwiftPM dependency is a `remoteSourceControl` pin (or a legacy `repositoryURL`, which this function also maps to Git). After this change, `bomly scan --enrich` on a SwiftPM project sends **zero** packages to OSV/Grype.

Unlike npm/Cargo — where a git pin genuinely isn't the published registry release — remote source control *is* the canonical distribution channel for Swift packages, and the OSV matcher's Swift support is keyed on exactly this: it queries the `SwiftURL` ecosystem, whose package identity is the repository URL (`internal/matchers/osv/matcher.go:605`). Excluding these pins as "not a registry release" inverts the intent for this ecosystem.

Suggestions (either works):
- Keep `Source = git` as truthful occurrence evidence, but carve SwiftPM out in the eligibility decision (e.g. `RegistryMatchEligible` or the detector could mark these occurrences eligible some other way), or
- Don't classify `remoteSourceControl`/legacy pins in this PR and land SwiftPM once the eligibility story is settled.

Whichever way, please add a test pinning the expected eligibility of a `remoteSourceControl` package so this can't regress silently — there's no enrich smoke case for SwiftPM to catch it.

internal/detectors/cargo/detector.go
```diff
@@ -372,20 +372,40 @@ func rootNode() *sdk.Dependency {
...
 		PackageManager: sdk.PackageManagerCargo,
 		Type:           sdk.ParsePackageType(pkgType),
 		Language:       "rust",
-		PURL:           sdk.BuildPackageURL("cargo", "", pkg.Name, pkg.Version)}, ResolvedURL: pkg.Source,
+		PURL:           sdk.BuildPackageURL("cargo", "", pkg.Name, pkg.Version)}, Source: source, ResolvedURL: pkg.Source,
 	})
 
 }
 
+func cargoDependencySource(source string) sdk.DependencySource {
```
bomly-guy, line 395: Two notes here:

1. **The workspace-from-lock path never uses this.** `depGraphFromLockWorkspace` (workspace.go:157) builds every node through its own local `nodeFor` closure, which sets no `Source` — so for a workspace resolved from `Cargo.lock` without the cargo binary, members aren't classified `workspace`/`project` and registry crates aren't classified `registry`, even though `parseCargoLockPackages` now extracts `pkg.Source`. That leaves the same project classifying differently depending on whether `cargo` is on `PATH`, and it contradicts the new doc text ("Bomly reads registry, Git, path, and workspace sources from Cargo metadata or `Cargo.lock`"). Worth wiring through in this PR since the parsed data is already there — or trimming the doc claim.

2. Minor: `strings.TrimSpace(source)` is evaluated four times; hoist it once at the top:

```go
func cargoDependencySource(source string) sdk.DependencySource {
	source = strings.TrimSpace(source)
	switch {
	case strings.HasPrefix(source, "registry+"), strings.HasPrefix(source, "sparse+"):
		return sdk.DependencySourceRegistry
	...
```

internal/detectors/python/uvlock.go
```diff
@@ -210,6 +215,39 @@ func depGraphFromUVLock(uvLockPath string) (*sdk.Graph, error) {
...
+func uvResolvedURL(source uvLockSource) string {
+	for _, value := range []string{source.Git, source.URL, source.Registry, source.Editable, source.Path} {
+		if value != "" {
+			return value
+		}
+	}
+	return ""
+}
+
+func uvSourceRevision(source uvLockSource) string {
```
bomly-guy, line 241: **[question]** Do `resolved` and `precise` (and a bare `rev`) actually appear as keys in real `uv.lock` source tables? As far as I can tell, uv encodes the requested ref as a query parameter and the resolved commit as the URL *fragment* — e.g. `source = { git = "https://github.com/owner/repo?rev=main#<commit-sha>" }` — rather than as separate TOML keys. If that's right, `uvSourceRevision` never fires on real lockfiles (the only test feeds it a synthetic struct), `source_revision` metadata stays empty for uv git deps, and the resolved commit ships embedded in `ResolvedURL` instead.

Could you verify against a `uv.lock` generated from a git dependency? If the fragment is the real carrier, parsing it out of `source.Git` would make the revision extraction actually do something; otherwise a comment citing the uv version that emits these keys would help the next reader.

internal/detectors/ruby/detector.go
```diff
@@ -448,13 +481,17 @@ func scopeForGroupLabels(labels []string) sdk.Scope {
 	return sdk.ScopeDevelopment
 }
 
-func gemNode(name, version string) *sdk.Dependency {
+func gemNode(spec lockSpec) *sdk.Dependency {
```
bomly-guy, line 484: Minor: this allocates a metadata map for every gem even when there's no revision, unlike the python path's `sourceRevisionMetadata`, which returns nil for the empty case. Same nil-when-empty shape keeps the two detectors consistent:

```go
var metadata map[string]any
if revision := strings.TrimSpace(spec.Revision); revision != "" {
	metadata = map[string]any{"source_revision": revision}
}
```

(Harmless today — `omitempty` drops empty maps and node metadata doesn't reach scan JSON anyway — purely a consistency/allocation nit.)

PR #387, opened 2026-08-13
feat: fully automated lockstep release train

bomly-guy:
Automates Bomly's lockstep component release train end to end: one `Auto Version` dispatch now versions, tags, pins, and releases the CLI **and every component module atomically** — no manual steps.

## Why a two-commit flow

The root `go.mod` must never carry `replace` directives, so that remote `go install github.com/bomly-dev/bomly-cli/cmd/bomly@latest` keeps working: every root tag has to carry pins that resolve on their own. But the pins can only resolve after the component tags exist, and the component tags have to describe the code being released — the classic pin/tag chicken-and-egg. The resolution is ordering, not replaces (think of it as Go's equivalent of Maven parent-version inheritance: the version is authoritative in one place, and automation propagates it to every module):

1. **Commit A** — version bump on `main` (`cmd/bomly/main.go`, npm wrapper, MCP Registry versions). No root tag yet.
2. **Component tags at A** — every `components/<kind>/<name>/vX.Y.Z`, pushed. Now `go get <module>@vX.Y.Z` resolves.
3. **Commit B** (`[skip ci]`) — the pin phase: `GOWORK=off go get` each component `@vX.Y.Z` + `go mod tidy`, touching only the root `go.mod`/`go.sum`.
4. **Root tag `vX.Y.Z` at B** — resolvable pins, zero replaces — then the existing `release.yml` dispatch.

**Lockstep invariant (tree identity):** component module zips contain only their own subtree, and commit B touches only the root `go.mod`/`go.sum`, so the `components/` tree is identical between A and B — the component tags describe exactly the code shipping in the CLI release. The workflow asserts `git diff --quiet A B -- components/` before tagging the root, and `verify` re-checks the invariant against origin.

## `scripts/release-components.sh` subcommands

Reworked into a small tool shared by CI and humans:

- `tag --version vX.Y.Z --commit <sha>` — annotated `components/<path>/vX.Y.Z` tags for every module in **that commit's tree** (`git ls-tree`; module directive validated from the commit, not the checkout). Remote-first idempotency: remote tag at commit A → skip; remote tag anywhere else → abort (never silently rides into a release); stale local-only tag → abort with a delete hint; local-only tag at the right commit (partial earlier run) → pushed.
- `pin --version vX.Y.Z` — workspace-off `go get <module>@vX.Y.Z` for every module in HEAD's tree + `go mod tidy`; no commit (the caller commits B). Defaults to `GOPROXY=direct` and skips the checksum DB for this repo's own component modules so freshly pushed tags resolve without proxy/sumdb lag (both overridable).
- `verify --version vX.Y.Z` — every component tag exists on origin and peels to a commit whose `components/` tree is identical to the root tag's.
- bare / `--version` — friendly dry run: prints what the `tag` and `pin` phases would do at the latest (or given) release tag, changing nothing (tree-identity-aware when reporting already-released tags). `make release-components` maps here.
- `--apply` is retired with a pointed error naming the replacement.

## Idempotency and recovery

Every phase of `auto-version.yml` is rerun-safe after a partial failure: commit A is skipped when already on `main`; the anchor commit A is re-derived from the remote component tags (so a rerun after commit B landed still tags at A, not B); existing tags at A are skipped; an already-applied pin skips commit B; the root-tag step re-pushes rather than re-creates. Manual recovery uses the same three subcommands; `dev-docs/RELEASE_CHECKLIST.md` documents them, including the failed-after-root-tag case (`gh workflow run release.yml --ref vX.Y.Z`).

**Zero-component compatibility:** with no `components/*/*/go.mod` in the tree, the tag and pin phases no-op with a log line and `verify` passes with 0 component tags — today's single-commit, single-tag behavior exactly.

## Validation

- `bash -n` clean; both workflows pass a YAML load.
- Scratch bare-repo rehearsal of the exact workflow command sequence against a workspace-design fake repo (committed `go.work`, workspace-only component, local `file://` GOPROXY standing in for the module proxy): full `v0.2.0` train (compute → bump → commit A → component tags at A → pin → commit B → tree-identity assert → root tag at B → verify), then: idempotent full rerun of every phase (commit A skipped, anchor re-derived from remote tags, pin no-op, verify green), local-tag-never-pushed recovery, remote-tag-at-wrong-commit abort (with verify correctly still passing while trees match, and correctly failing once the `components/` tree actually drifts), stale-local-tag abort + recovery, missing-tag verify failure with recovery hint, both dry-run forms, and a zero-component `v0.3.0` release. All checks passed; `GOWORK=off go build` resolves the pins at commit B.
- No Go code changes; `make test` is unaffected.

## Notes

- Supersedes the manual release-train step described in #386's flow (`--apply` + follow-up pin-bump PR): tagging and pinning are now inside `auto-version.yml`, and the pin lands as the automated commit B rather than a PR.
- `ci.yml`'s pinned-build/tidy skip notices now point at the automation instead of the manual steps.

### DIFF
--- a/.github/workflows/auto-version.yml
+++ b/.github/workflows/auto-version.yml
@@ -101,10 +101,18 @@ jobs:
           echo "version=${next}" >> "$GITHUB_OUTPUT"
           echo "tag=${tag}" >> "$GITHUB_OUTPUT"
 
+      - name: Set up Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: go.mod
+          cache: true
+          cache-dependency-path: go.sum
+
       - name: Update source version
         shell: bash
         env:
           VERSION: ${{ steps.version.outputs.version }}
+          TAG: ${{ steps.version.outputs.tag }}
         run: |
           set -euo pipefail
           sed -i -E "s/^var version = \"[^\"]+\"/var version = \"${VERSION}\"/" cmd/bomly/main.go
@@ -119,7 +127,7 @@ jobs:
 
           git diff -- cmd/bomly/main.go npm/package.json server.json
 
-      - name: Commit version bump and tag
+      - name: Commit version bump (commit A)
         shell: bash
         env:
           TAG: ${{ steps.version.outputs.tag }}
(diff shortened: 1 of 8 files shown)


bomly-guy:
Closing unmerged: the lockstep release train this PR automates is obsolete. The component-extraction program changed direction from in-repo nested modules to external per-component repositories (`bomly-plugin-*`), consumed as ordinary pinned Go modules and bumped by Dependabot — there are no in-repo component modules left to tag, so there is no release train to run.

See https://github.com/bomly-dev/bomly-cli/pull/389 for the consuming change and the recorded architecture decision.

PR #439, opened 2026-09-06
feat(detectors): record which module root each location belongs to (phase 2.8, producer half)

bomly-guy:
Producer half of SDK maturity plan row 2.8. Native detectors now record, on every `PackageLocation` they emit, the module root whose resolution produced the site, that site's relationship, and that site's scopes — the join key ADR-0037 ("Usage facts carry their attribution") defines as `(module root, declaration site)`.

## What attribution now lands

One shared helper, `detectors.Attributed` (`internal/detectors/attribution.go`), is the only place the derivation lives; every detector returns its `sdk.DetectionResult` through it, and a guard test (`TestDetectionResultsCarryingGraphsAreAttributed`) fails when one does not.

- **Module root** — the directory of the module node's declaring manifest (`.` for the detector's own working directory, `apps/web` for a workspace member, `module-a` for a Maven reactor project).
- **Relationship** — derived from that root's own edges: direct when the root declares it, transitive when reached through another dependency, and `unknown` preserved when the detector already said the parent could not be recovered. Deliberately *not* `sdk.RelationshipForPath`, which prefers the node's stored scalar — the merged value that cannot tell one root's declaration from another root's transitive path. That decline is recorded in the code.
- **Scopes** — precise when the detector supplies per-module declarations (npm does, via `ModuleDeclarations`); otherwise the node's union, but only when exactly one module root reaches the node. When several do, the union mixes roots and the site claims *no* scope rather than a scope nobody observed there.
- **Ownership of a site** — a file inside a module's own directory belongs to that module alone (a Maven sibling does not claim `module-a/pom.xml`); a file at the top of the scan is shared, so each workspace member that reaches a lockfile line gets its own record of it.

`internal/engine/consolidation` rebases the module root next to the paths it already rebases: every detector calls its own working directory `.`, and two subprojects both claiming `.` would make the key useless the moment a recursive scan merges them. That is the only engine change; the new fields otherwise survive consolidation untouched (verified).

## The row 2.8 regression case

`TestNPMLockfileWorkspaceRecordsOneLocationPerModuleUsage` (new fixture `npm-v3-workspaces-cross-scope`): `shared-tool` is a **direct development** dependency of `apps/web` and a **transitive runtime** dependency of `packages/lib`. It comes back as two records of the same lockfile line — `{apps/web, direct, [development]}` and `{packages/lib, transitive, [runtime]}` — while the node's own scope union keeps both, unchanged.

`TestMavenSharedDependencyGetsOneRecordPerReactorModule` is the reactor form: one artifact, one record per reactor module, each pointing at its own pom, and no invented scope because Maven passes no declarations.

## Which detectors attribute, and which cannot

Attributing (module root + relationship on every location they emit): cargo, cocoapods, composer, conan, go, gradle, maven, mix, npm, nuget, pnpm, poetry, pip, pipenv, pub, ruby, sbt, swiftpm, uv, yarn.

Not attributing, exemption recorded in the file and named by the guard test:

- **sbom** — it converts a document, it does not resolve a project. Whatever module produced those packages was resolved elsewhere by something else; the paths are the producer's, not this scan's.
- **github-actions** — a workflow is not a module. Its root is the workflow file, so the only available "module root" would be `.github/workflows`, a directory that declares no module.
- **bun** — attaches no positions, so it emits no locations for attribution to land on. Nothing to attribute until it does.

Scope precision beyond the module root is currently supplied only by npm; pnpm, yarn, cargo and maven have per-module declarations available and could be upgraded the same way in a follow-up.

## Golden movement — read this before merging

`make verify` passes with **no** drift: `internal/output.LocationRef` projects only `real_path`, `access_path` and `position`, so single-module scan JSON is byte-identical.

Two smoke goldens **do** move, and I did **not** regenerate them: `scan-npm-workspaces` and `scan-cargo-workspace`. Because the projection drops the attribution, the shared-lockfile records for two workspace members render as *duplicate* location entries — a diff that shows something other than added attribution, which is a bug rather than drift. The fix belongs to the consumer half of row 2.8 (project `module_root` / `scopes` / `relationship` into `LocationRef`), which is out of this PR's scope.

**So: this PR should land together with, or after, the consumer-half change that projects the attribution, and those two goldens should be regenerated then.** Alone it leaves those two smoke cases red and workspace scan JSON showing duplicated-looking locations.

Related SDK gap found while verifying (recorded in the code, worth an SDK issue): `hasDependencyLocation` compares paths and position only, so merging two *separate* nodes that each carry one root's record drops the second. Workspace detectors partition with shared node pointers, so the case above survives; two detectors resolving one package from one path would not.

## Mutations run

Each applied alone, tree kept compiling, then reverted:

| # | Mutation | Result |
|---|---|---|
| M1 | direct children recorded as transitive | fails `TestAttributedRecordsTheModuleRootAndDirectness`, npm, maven, gomod |
| M2 | site scopes always empty | fails `TestAttributedGivesEachModuleRootItsOwnRecord`, cycle test, npm |
| M3 | never add a second record for a shared site | fails both npm workspace tests and both multi-root helper tests |
| M4 | every root claims every site | fails `TestAttributedLeavesAnotherModulesFileAlone`, maven |
| M5 | module root not rebased for a subproject | fails `TestRebaseGraphLocations_MovesModuleRootIntoRepositoryCoordinates` |
| M6 | `Attributed` does nothing | fails all 8 helper tests, all 13 detector-table cases, npm, maven, gomod |
| M7 | (first attempt) skip the `Name` lookup only | passed — a non-mutation: `QualifiedName()` catches it. Redone as M7b |
| M7b | declared scopes never consulted | fails `TestAttributedGivesEachModuleRootItsOwnRecord`, cycle test, npm |
| M8 | claim the node union even when several roots reach it | fails `TestAttributedLeavesScopesEmptyWhenTheUnionMixesModules`, maven |
| M9 | module roots normalized to a wrong directory | fails every attribution test |
| M10 | one detector's wrapper removed (parenthesized form) | fails the detector table; the guard regex does not match this form |
| M11 | one detector forgets `Attributed` (the realistic form) | fails `TestDetectionResultsCarryingGraphsAreAttributed` |

## Delegation check

Ran the `delegation-check` skill before implementing. Outcome: no new grammar, vocabulary, or format is introduced. The scope lattice, relationship vocabulary and merge rules are the SDK's and are used through it (`sdk.MergeScope`, `sdk.ScopesOf`, `sdk.DependencyRelationship*`, `sdk.AsDependencyNode`); path handling is `path`/`strings`. One decline is recorded in the code: `sdk.RelationshipForPath` is not used for per-site directness because it prefers the stored merged scalar, which is exactly the lossy summary this change exists to replace. No parser added, so no new fuzz target is required.


bomly-guy:
**Follow-up pushed: this PR is now self-contained and green.**

It was opened deliberately red, with two workspace smoke goldens left unregenerated because the output projection dropped the new attribution — so a workspace reaching one lockfile line from two members rendered the same path twice, byte-identical. That was the right call: it is a regression, not drift, and blessing it would have shipped output worse than the summary it replaced.

Rather than leave a merge-ordering dependency, I carried the attribution into `output.LocationRef` here:

```
 "access_path": "package-lock.json",
+"module_root": "apps/web",
 "real_path": "package-lock.json",
+"relationship": "direct",
+"scopes": ["development"]
```

Those two records are now distinguishable, and a conjunctive question — reachable, runtime and direct — can be asked of one usage instead of three summarized onto a package.

**Golden movement was wider than the two predicted**: 36 goldens carry locations, including the audit, diff, explain and lite suites. Audited rather than accepted — across all 36 the only added keys are `module_root`, `relationship`, `scopes` and the paths of records that now appear distinctly, and **not one removed line is anything but a `real_path`/`access_path` re-emitted with a trailing comma**.

`finding-baseline-workflow` moved too, and I checked that one separately because it is the compatibility question the plan calls out: finding ids, dependency refs, package identity and the audit summary are all untouched. Baselines still key on package and finding references, not locations.

Schemas regenerated (`scan`, `diff`, `explain`). `LocationRef` loses comparability by carrying a slice — the same deliberate break the SDK took on `PackageLocation`; nothing compares one or keys a map by one, and the type now says so.

Mutations: dropping the attribution from the projection fails `TestLocationRefsCarryTheAttributionThatDistinguishesThem`; dropping an attributed site with no path fails `TestAnAttributedSiteWithNoPathIsKept`.

`make verify SMOKE=1` green.

